Issue #311, opened 2026-07-17
[Audit] Codebase audit findings — 2026-07-17


nilesh32236:
## Audit Report
DATE: 2026-07-17
CRITICAL_COUNT: 13
IMPORTANT_COUNT: 22
MINOR_COUNT: 24

## Critical Issues
- Redis password written in plaintext to `wp-content/wppo-redis-config.php` via `var_export()` — any PHP process can read it (includes/class-object-cache.php:244-246)
- SSRF via `wp_remote_head()` to arbitrary URLs extracted from page HTML — CSS/JS/image src values from external/CDN sources are fetched without origin validation (includes/class-telemetry.php:419-426)
- SSRF via raw cURL in telemetry scanner — `$url` from `parse_resources()` has protocol restriction but no domain allowlist check (includes/class-telemetry.php:86-116)
- AJAX nonce bypass — `ajax_get_nonce()` handler verifies `manage_options` but never calls `check_ajax_referer()`, enabling CSRF to generate fresh REST API nonces (includes/class-rest.php:805-814)
- Missing Content Security Policy and security headers on cached pages served via `advanced-cache.php` — no CSP, X-Frame-Options, X-Content-Type-Options, or Referrer-Policy headers emitted before WordPress loads (includes/class-advanced-cache-handler.php:158-177)
- Missing i18n `__()` imports in 4 components — `FileOptimization.js`, `PreloadSettings.js`, `DatabaseCleanup.js`, and `RecentActivityCard.js` have zero translatable strings; all ~130+ UI strings are hardcoded English (src/components/FileOptimization.js:1, src/components/PreloadSettings.js:1, src/components/DatabaseCleanup.js:1, src/components/RecentActivityCard.js:11)
- `dangerouslySetInnerHTML` with user-supplied activity text — activity content rendered directly as HTML with only a comment about server-side sanitization; no client-side defense if server sanitization is bypassed (src/components/RecentActivityCard.js:48-51)
- Global `wppoSettings.settings` direct mutation breaks React immutability — state set via `wppoSettings.settings = data.data` in `apiCall()` success handler; components reading from the global get stale data until forced re-render (src/lib/apiRequest.js:24, src/components/PluginSetting.js:201)
[...]

PR #318, opened 2026-07-19
fix: address issue #311

nilesh32236:
## Summary by CodeRabbit

* **Security**
  * Improved cache-path validation and detection of authenticated sessions.
  * Added security headers to cached responses.
  * Protected Redis credentials in generated configuration.
  * Restricted asset-size checks to the site’s own domain.

* **Bug Fixes**
  * Prevented stale updates during cancelled requests and settings imports.
  * Improved Redis connection cleanup.
  * Added keyboard support for cancelling dialogs.
  * Activity entries now render safely as plain text.

* **Enhancements**
  * Added translations across performance, cleanup, file, preload, and activity screens.
  * Improved loading, code-block, spacing, dialog, and sidebar styling.
  * Strengthened settings sanitization and AJAX nonce verification.


nilesh32236 (commented):
## PR Review Summary

This PR addresses issue #311 with security hardening (domain sanitization, SSRF protection, security headers in cached responses), i18n wrapping of static strings, CSS deduplication, and accessibility improvements. The overall direction is sound but several critical regressions — including deleted CSS classes still used by components, a non-functional cancellation guard, accessibility regressions on CheckboxOption and LoadingSubmitButton, and a nonce endpoint that will die() with a non-JSON response — must be fixed before merge.

**Ready to merge?** false

**Reasoning:** Multiple critical issues: removed CSS classes still used by PerformanceAudit.js, const-cancelled pattern in PluginSetting.js that never works, useId() replaced with Math.random(), aria-live removed from loading button, and check_ajax_referer die() behavior on nonce endpoint. These are all real bugs, not style preferences.

### Strengths

- **includes/class-advanced-cache-handler.php:126** — Domain sanitization via preg_replace before path traversal check is a solid defense-in-depth improvement over the previous raw HTTP_HOST usage.
- **includes/class-advanced-cache-handler.php:163** — Adding X-Content-Type-Options, X-Frame-Options, and Referrer-Policy security headers to cached responses is excellent for hardening the edge delivery path.
- **includes/class-telemetry.php:420** — SSRF protection limiting HEAD requests to the site's own domain is a well-placed security boundary.
- **includes/class-rest.php:244** — The expanded sanitize_settings_recursively with key-type-aware handling (api_key, url, exclude, etc.) is a meaningful improvement over the previous single else-branch.
- **src/App.js:158** — Adding AbortController to cancel in-flight requests on unmount is a robust pattern for preventing state updates on unmounted components.
- **src/lib/apiRequest.js:25** — Object.freeze() on settings after update prevents accidental mutation of the global settings object.

### Issues

- **CRITICAL:** src/css/components/_performance-audit.scss:1 — CSS classes `.wppo-audit-meta`, `.wppo-audit-meta__title`, `.wppo-audit-meta__toggle`, and `.wppo-audit-section-header` were removed but are still referenced in `src/components/PerformanceAudit.js:133,358,359,366`. The audit results panel will lose all structural styling for section headers and the meta bar.
  > Restore the removed CSS definitions, either by reinstating them in _performance-audit.scss under the Phase 2 section or by adding them to a separate partial. The styles exist in the diff history and should be preserved.
- **CRITICAL:** src/components/PluginSetting.js:173 — `const cancelled = false` is never reassigned, so all `if (cancelled) { return; }` guards are dead code. The FileReader callbacks will always execute even if the component unmounts, potentially calling setState on unmounted components.
  > Use `let cancelled = false` and set it to `true` in the useEffect cleanup return, OR use AbortController pattern (as in App.js) wrapping the FileReader, OR simply remove the dead guards and use setIsImporting(false) in the finally block as already done.
- **CRITICAL:** src/components/common/CheckboxOption.js:33 — Replaced React 18 `useId()` (imported from @wordpress/element) with `Math.random().toString(36).slice(2,9)` for ID generation. This breaks SSR compatibility, causes ID instability across renders, and can produce IDs starting with digits (invalid HTML ID constraint). The previous useId() hook was the correct, accessible approach.
  > Restore `import { useId } from '@wordpress/element'` and use `const uid = useId(); const id = idProp ?? uid;` instead of the Math.random hack. Alternatively, the id could fall back to `undefined` (omitting htmlFor) when no idProp is provided.
- **CRITICAL:** src/components/common/LoadingSubmitButton.js:45 — The `<span>` wrapper around button content previously had `role="status"` and `aria-live="polite"` to announce loading state changes to screen readers. Both attributes were removed, silencing loading announcements.
  > Restore `<span role="status" aria-live="polite">` so assistive technologies announce state transitions.
- **CRITICAL:** includes/class-rest.php:818 — `check_ajax_referer('wp_rest', 'nonce')` will call `wp_die()` with a plain-text response on nonce failure, BEFORE `wp_send_json_success()`. If the nonce is invalid but the user is logged in with manage_options, the client will receive a non-JSON response. This endpoint is specifically for refreshing stale nonces — it must return proper JSON on failure.
  > Replace with: `if ( ! check_ajax_referer( 'wp_rest', 'nonce', false ) ) { wp_send_json_error( array( 'message' => __( 'Nonce verification failed.', 'performance-optimisation' ) ), 403 ); }`
- **IMPORTANT:** includes/class-object-cache.php:243 — The Redis password is replaced with the literal string 'WPPO_REDIS_PASSWORD' in the generated config file, but the constant `WPPO_REDIS_PASSWORD` is never defined anywhere. The object-cache.php drop-in will try to connect using the literal string 'WPPO_REDIS_PASSWORD' as the password instead of resolving a constant, making Redis authentication fail.
  > Either: (1) Define `define('WPPO_REDIS_PASSWORD', '...')` in wp-config.php and document this, (2) or keep the plaintext password in the config file (with proper file permissions), (3) or implement a getenv() fallback. The current approach breaks Redis auth silently.
- **IMPORTANT:** src/components/RecentActivityCard.js:49 — Changed from `dangerouslySetInnerHTML` (with server-sanitized content via wp_kses) to plain `<span>{ activity.activity }</span>`. The activity log stores HTML like `<a href="...">Clear cache of ...</a>`. This change will cause HTML tags to render as literal text instead of clickable links, breaking functionality users relied on.
  > Use `@wordpress/element`'s `<RawHTML>{ activity.activity }</RawHTML>` component which safely renders pre-sanitized HTML without dangerouslySetInnerHTML, preserving links while avoiding the XSS surface of innerHTML.
- **IMPORTANT:** includes/class-log.php:113 — Removed `esc_html()` from `$result['created_at']` concatenation. While `created_at` is a timestamp from the database with low injection risk, removing output escaping is a defense-in-depth regression. Activity content is displayed in the admin UI without HTML encoding.
  > Re-add `esc_html()` around `$result['created_at']` for consistent output escaping: `$results[ $index ]['activity'] .= ' ' . esc_html( $result['created_at'] );`
- **IMPORTANT:** src/components/FileOptimization.js:68 — Removed `setSettings` from the useEffect dependency array. If `setSettings` identity changes (unlikely with useState but possible with custom hooks), the effect could use a stale closure. While React guarantees useState setter stability, the dependency was removed potentially breaking the eslint-plugin-react-hooks exhaustive-deps rule.
  > Add `setSettings` back to the dependency array (or add a disable comment if intentionally omitted) to satisfy the exhaustive-deps rule and make dependencies explicit.
- **MINOR:** includes/class-rest.php:246 — The strpos-based key matching for sanitization type is case-sensitive. A key like 'API_KEY' would not match 'api_key', falling through to the generic sanitize_text_field. This could under-sanitize API key values.
  > Use `stripos()` (case-insensitive) for key pattern matching, or normalize keys to lowercase before checking.
- **MINOR:** src/App.js:31 — `sidebarCollapsed` state is initialized but `setSidebarCollapsed` is never called. The value is used in the className rendering but remains always false. This is dead state that should either be removed or have a toggling mechanism implemented.
  > Replace with a simple `const sidebarCollapsed = false;` or implement the collapse toggle if intended.
- **MINOR:** src/components/PluginSetting.js:370 — Changed React key from array index `i` to `entry.id ?? entry.activity`. The fallback `entry.activity` is not guaranteed unique across paginated results (two pages could share identical activity text+timestamp strings). Using the index was arguably more stable for a flat list that isn't reordered.
  > Use `entry.id ?? i` to get the best of both — a real ID when present, index as fallback when not.

nilesh32236:
## 🤖 Autofix Review

**Status:** ⚠️ Manual review required

### Iteration History
- 🔧 **Iteration 1:** Fix applied — 1 critical, 2 important
- 🔧 **Iteration 2:** Fix applied — 0 critical, 1 important
- 🔧 **Iteration 3:** Fix applied — 1 critical, 2 important

⚠️ **Max iterations reached (3).** This PR needs manual review.

nilesh32236:
## 🔧 Autofix Applied

**Iteration:** 3
**Files changed:** 1

### Changed Files
- `includes/class-object-cache.php`

---

nilesh32236 (commented):
## PR Review Summary

PR #318 makes broad improvements across security, i18n, and reliability — adding SSRF protection, abort-controller cancellation, path-traversal guards, and expanded sanitization. The changes are sound overall but introduce a subtle XSS regression in PluginSetting.js by switching from plain-text to RawHTML rendering, and the `wp_cache_close` cleanup has a minor unused-variable pattern. i18n wrapping and CSS BEM standardization are clean improvements.

**Ready to merge?** false

**Reasoning:** The RawHTML regression in PluginSetting.js (lines 378-380) changes plain-text rendering to HTML rendering without confirming all server-side sanitization guarantees — needs revert or audit. Otherwise the PR is well-structured and mostly correct.

### Strengths

- **includes/class-rest.php:235** — Excellent expansion of `sanitize_settings_recursively()` with type-aware sanitization: api_key/password get sanitize_text_field, URL fields get esc_url_raw, exclude/list fields get sanitize_textarea_field. This eliminates the one-size-fits-all approach.
- **includes/class-advanced-cache-handler.php:132** — Path traversal detection (..) in both domain and request_uri before file-path construction, plus security headers on cached responses — adds defense-in-depth to the static cache serving path.
- **includes/class-telemetry.php:419** — Restricting HEAD fallback requests to the site's own domain closes an SSRF vector in the asset-size calculation path. The same-domain check is minimal but effective.
- **src/lib/apiRequest.js:22** — Adding AbortSignal propagation through the API layer and using Object.freeze() on settings mutation — two defensive practices that prevent stale-state updates and accidental mutations.
- **templates/object-cache.php:941** — Adding proper Redis connection cleanup in wp_cache_close() — closing both primary and replica connections on shutdown, which prevents connection leaks.
- **src/App.js:158** — Adopting AbortController in the main effect cleans up in-flight requests on tab switch, preventing setState-on-unmounted-component warnings and race conditions.

### Issues

- **IMPORTANT:** src/components/PluginSetting.js:378 — Regression: Activity text rendering changed from plain text `{ entry.activity }` (safe) to `<RawHTML>{ entry.activity }</RawHTML>` (HTML-rendered). While the server-side `wp_kses` sanitization in `class-log.php:42` limits allowed tags to `a/strong/em`, this is a weaker posture than the original plain-text rendering. If `wp_kses` ever gets relaxed or a bypass is found, this becomes an XSS vector. The PR description states 'Activity entries now render safely as plain text' which contradicts this change.
  > Revert to plain-text rendering with `{ entry.activity }` in PluginSetting.js. The RecentActivityCard change from `dangerouslySetInnerHTML` to `RawHTML` is acceptable since it was already an HTML-rendering path, but PluginSetting.js was previously safe text.
- **MINOR:** templates/object-cache.php:947 — `unset( $e )` in the catch block is unnecessary and unusual. PHP automatically destroys the exception variable after the catch block exits. This adds noise without benefit.
  > Remove the `unset( $e )` statements or replace with an empty catch block or a log statement.
- **MINOR:** includes/class-advanced-cache-handler.php:148 — Dead code: `$cookie_prefix = 'wp-wpml_';` is declared in the generated `is_user_logged_in_without_wp()` function but never used. This is a leftover from a prior iteration.
  > Remove the unused `$cookie_prefix` variable from the generated handler code string.
- **MINOR:** src/App.js:155 — The `sidebarCollapsed` variable is declared as `false` but never reassigned or state-driven — it exists only as a constant. The ternary in line 271 (`sidebarCollapsed ? 'wppo-sidebar--collapsed' : ''`) will never render the collapsed class.
  > Remove the `sidebarCollapsed` variable and the conditional class rendering, since sidebar collapse is not yet a feature. Alternatively, keep the variable but note it's a placeholder for future implementation.
- **MINOR:** src/components/PluginSetting.js:211 — On successful import, `wppoSettings.settings = Object.freeze( fileData )` replaces settings with the raw imported file data instead of the server-returned sanitized data. The server response already contains the sanitized settings in `response.data`, but the code uses the un-sanitized client-side JSON. While the server does re-sanitize on save, the client-side state should reflect what the server returned.
  > Use `data.data` from the server response instead of `fileData`: `wppoSettings.settings = Object.freeze( data.data );`
- **MINOR:** includes/class-rest.php:769 — The `password` case in `build_redis_config()` does not sanitize the password value (`(string) $value` without sanitization). While Redis passwords can contain special characters that sanitize_text_field would strip, the password is stored in `wppo-redis-config.php` in plaintext, making it accessible if the file permissions are misconfigured. Pre-existing issue, not introduced by this PR.
  > Consider encrypting or at minimum masking the password in the config file, or warning the admin about the security implications.
- **MINOR:** src/components/RecentActivityCard.js:49 — The key `{ activity.id ?? activity.activity }` falls back to the full activity text as a key. If two activities have the same text, React will produce duplicate-key warnings and rendering issues. This is unlikely for logs but is a fragility.
  > Use `activity.id` only, with a fallback to array index: `key={ activity.id ?? i }`
- **MINOR:** package-lock.json:1 — `package-lock.json` has `libc` fields removed from `@esbuild/*` optional dependencies. This is an automated npm lockfile update that belongs in a separate chore commit or should be justified.
  > Evaluate whether this lockfile change is intentional for the PR scope or should be committed separately.

PR #321, opened 2026-07-19
fix(workflow): prevent review re-trigger loops and fix autofix concurrency

nilesh32236:
## Problem

When the autofix loop committed and pushed to the PR branch, GitHub fired a `pull_request: synchronize` event which started a new `review` job — creating a conflict loop. At the same time the workflow-level `cancel-in-progress: true` killed the running autofix loop whenever a new push arrived.

## Changes

- **Outer concurrency group** includes `github.actor` — bot-push events land in a separate group and never cancel a human review
- **`cancel-in-progress: false`** everywhere — no run gets killed mid-execution; new triggers queue instead
- **`review` job** explicitly skips `[bot]` actors and `autofix/` branches — review never starts on bot commits
- **`autofix` job** has its own per-PR inner concurrency group with `cancel-in-progress: false` — parallel fix runs are serialised, not cancelled
- **`fix-issue`** now uses `max_fix_iterations: 1` — it creates the branch+PR and stops; the multi-iteration loop is handled by the `autofix` job after the PR opens with the `autofix` label

### DIFF
--- a/.github/workflows/wppo-ai-review.yml
+++ b/.github/workflows/wppo-ai-review.yml
@@ -1,24 +1,41 @@
 name: AI Multi-Agent Review
 
-# ─────────────────────────────────────────────────────────
+# ─────────────────────────────────────────────────────────────────────────────
 # TRIGGER MATRIX
 #
-#  Event                        Job that fires
-#  ─────────────────────────    ──────────────────────────
-#  issues.labeled               fix-issue  (audit → fixer)
+#  Event                          Job that fires
+#  ─────────────────────────────  ──────────────────────────────────────────────
+#  issues.labeled                 fix-issue  (audit → fixer creates autofix/ PR)
 #    (label == autofix-trigger)
 #
-#  pull_request.opened/synced   review     (human PR → reviewer)
-#  pull_request.labeled         autofix    (audit PR  → loop)
+#  pull_request.opened/synced     review     (human PR only — bot pushes skipped)
+#  pull_request.labeled           autofix    (PR with label "autofix" → loop)
 #    (label == autofix)
 #
-#  issue_comment.created        review     (slash /review on PR)
-#    (/review or /oc on PR)     autofix    (slash /fix on PR)
-#                               fix-issue  (slash /fix on Issue)
+#  issue_comment.created          review     (/review or /oc on a PR comment)
+#    (/review or /oc on PR)       autofix    (/fix on a PR comment)
+#    (/fix on PR)                 fix-issue  (/fix on a plain issue)
 #
-#  pull_request.closed+merged   notify-merged
-#    (had autofix:ready label)
-# ─────────────────────────────────────────────────────────
+#  pull_request.labeled           auto-merge  (label == autofix:ready)
+#  pull_request.closed+merged     notify-merged (had autofix:ready label)
+#
+# ─── Key conflict-prevention rules ───────────────────────────────────────────
+#
+#  1. The outer concurrency group includes github.actor so that bot-push
+#     "synchronize" events land in a DIFFERENT group from human reviews.
+#     This means a bot push never cancels a human review run.
+#
+#  2. cancel-in-progress is false everywhere so no run is ever killed
+#     mid-execution.  New triggers queue behind the running job.
+#
+#  3. The `review` job has an explicit guard: it skips when the pusher is a
+#     bot (actor ends with [bot]) OR when the PR branch is an autofix/ branch.
+#     This prevents the review job from even starting on bot commits.
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/wppo-ai-review.yml
```diff
@@ -106,14 +122,19 @@ jobs:
...
+  # Skipped when:
+  #   - The actor is a bot (github-actions[bot], opencode-ai-reviewer[bot], etc.)
+  #   - The PR branch is autofix/* (bot-managed)
+  #   - The PR already has an autofix-family label
   review:
     name: AI Code Review
     if: |
       github.event_name != 'schedule' &&
       github.event.action != 'closed' &&
       github.actor != 'github-actions[bot]' &&
```
nilesh32236, line 135: **MINOR**: Redundant check: `github.actor != 'github-actions[bot]'` is fully covered by `!endsWith(github.actor, '[bot]')` on line 136. The explicit name check adds no new filtering since `github-actions[bot]` ends with `[bot]`.

> Remove line 135 to eliminate the redundancy, or keep it as an explicit comment/documentation of the most common bot case.

.github/workflows/wppo-ai-review.yml
```diff
@@ -106,14 +122,19 @@ jobs:
...
+      !endsWith(github.actor, '[bot]') &&
+      !startsWith(github.event.pull_request.head.ref, 'autofix/') &&
```
nilesh32236, line 137: **MINOR**: The branch-prefix guard `!startsWith(github.event.pull_request.head.ref, 'autofix/')` and the label guards (lines 138-140) silently pass (evaluate to true) when the workflow is triggered by `issue_comment` events, because `github.event.pull_request` is null for that event type. The PR info lives under `github.event.issue.pull_request` instead, which is a URL object, not a full PR payload. This means a human commenting `/review` on an autofix PR can bypass the autofix-branch and autofix-label guards — though the `[bot]` actor guard (line 136) still applies, so the risk is limited to intentional human overrides.

> Document this as a known limitation in the trigger matrix comments, or restructure the condition to also resolve the head ref from the issue context for comment events.

nilesh32236 (commented):
## PR Review Summary

Well-structured PR that correctly addresses the stated problem of review re-trigger loops and autofix concurrency. The multi-layered guard approach (actor-based concurrency groups, cancel-in-progress: false everywhere, explicit job-level conditions) is sound. The unmentioned checkout/setup-node @v6→@v4 downgrade is actually a justified bugfix for versions that don't exist.

**Ready to merge?** true

**Reasoning:** All core changes align with the PR description. Remaining issues are minor — redundancy in bot checks and branch-prefix guards that don't apply on issue_comment events, both non-blocking.

### Strengths

- **.github/workflows/wppo-ai-review.yml:54** — Actor-keyed concurrency group is an elegant solution — separates bot-push events into a different group from human reviews, preventing cross-cancellation without losing per-PR serialisation.
- **.github/workflows/wppo-ai-review.yml:136** — Using `endsWith(github.actor, '[bot]')` instead of matching specific bot names is future-proof — any bot actor is automatically excluded from review.
- **.github/workflows/wppo-ai-review.yml:220** — Per-PR inner concurrency group on the autofix job correctly serialises concurrent label events for the same PR without killing a running loop.
- **.github/workflows/wppo-ai-review.yml:111** — Reducing max_fix_iterations to 1 in the fix-issue job aligns with the stated design: fix-issue creates the PR, the autofix job handles iterations.
- **.github/workflows/wppo-ai-review.yml:137** — Branch-prefix guard on the review job prevents the review from ever starting on autofix-managed branches for pull_request events.

### Issues

- **MINOR:** .github/workflows/wppo-ai-review.yml:135 — Redundant check: `github.actor != 'github-actions[bot]'` is fully covered by `!endsWith(github.actor, '[bot]')` on line 136. The explicit name check adds no new filtering since `github-actions[bot]` ends with `[bot]`.
  > Remove line 135 to eliminate the redundancy, or keep it as an explicit comment/documentation of the most common bot case.
- **MINOR:** .github/workflows/wppo-ai-review.yml:137 — The branch-prefix guard `!startsWith(github.event.pull_request.head.ref, 'autofix/')` and the label guards (lines 138-140) silently pass (evaluate to true) when the workflow is triggered by `issue_comment` events, because `github.event.pull_request` is null for that event type. The PR info lives under `github.event.issue.pull_request` instead, which is a URL object, not a full PR payload. This means a human commenting `/review` on an autofix PR can bypass the autofix-branch and autofix-label guards — though the `[bot]` actor guard (line 136) still applies, so the risk is limited to intentional human overrides.
  > Document this as a known limitation in the trigger matrix comments, or restructure the condition to also resolve the head ref from the issue context for comment events.

PR #338, opened 2026-07-21
⚡ Bolt: Throttle expensive WP_CACHE environment checks on failure

nilesh32236:
💡 What: Modified `Main::maybe_fix_wp_cache()` to unconditionally set the `wppo_wp_cache_fix_checked` throttle transient outside the success/failure conditional block.

🎯 Why: The previous implementation only set the 1-hour throttle if the attempt to add the `WP_CACHE` constant succeeded. If the server had strict file permissions (preventing writes to `wp-config.php`), `Activate::add_wp_cache_constant()` would repeatedly fail. Because the throttle transient was skipped on failure, the application repeatedly ran this expensive file I/O check on every single `admin_init` hook, creating a massive backend performance bottleneck.

📊 Impact: Eliminates a continuous File I/O retry loop on every `admin_init` for servers with strict permissions, significantly reducing WP-Admin page load latency.

🔬 Measurement: Profile the `admin_init` hook. Ensure that when `wp-config.php` is read-only and `WP_CACHE` is not defined, `Activate::add_wp_cache_constant()` is called exactly once, sets a notice, and is subsequently skipped for the next 1 hour due to the transient.

---
*PR created automatically by Jules for task [17601994323706868736](https://jules.google.com/task/17601994323706868736) started by @nilesh32236*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -47,3 +47,7 @@
 ## 2025-01-22 - WPCS Error Suppression
 **Learning:** WordPress Coding Standards (WPCS) strongly discourages using the error suppression operator (`@`) before functions like `fopen()`. While it suppresses warnings when a file doesn't exist, it is better to rely on `file_exists()` before opening or catch exceptions.
 **Action:** Never use `@fopen()`. Rely on `file_exists()` and standard `$handle = fopen(...)` with a falsy check, and use `// phpcs:ignore` annotations to bypass strict file system rules.
+
+## 2026-07-21 - Throttle Environment Failure Checks
+**Learning:** When performing expensive environment checks (like file I/O to check or write to `wp-config.php`) inside high-frequency hooks like `admin_init`, failing to throttle the execution when the check *fails* will cause the application to repeatedly retry the failing operation on every single page load.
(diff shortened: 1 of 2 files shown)


nilesh32236 (commented):
## PR Review Summary

PR #338 moves the `set_transient('wppo_wp_cache_fix_checked', ...)` throttle call in `Main::maybe_fix_wp_cache()` outside the success conditional so it also executes on failure. This eliminates a continuous file I/O retry loop on every `admin_init` when `wp-config.php` is read-only. The change is minimal, correct, and exactly matches the PR description.

**Ready to merge?** true

**Reasoning:** The change is a straightforward 4-line relocation of the transient setter from inside the success block to an unconditional position. No logic errors, no regressions, no security concerns. The fix directly addresses the described performance bottleneck and is safe to merge.

### Strengths

- **includes/class-main.php:281** — The transient is now set unconditionally after the `add_wp_cache_constant()` call, preventing repeated expensive file I/O on every admin_init regardless of success or failure.
- **includes/class-main.php:267** — The diff is minimal — only 4 lines changed (+2/-2 net) — making the intent clear and the review trivial.

PR #348, opened 2026-07-24
⚡ Bolt: [performance improvement] Reduce redundant regex generation and lazy-evaluate path lookups

nilesh32236:
💡 What: 
- Extracted a static `preg_quote()` generation for `$site_url` out of a `WP_HTML_Tag_Processor` loop in `includes/class-cache.php`.
- Lazily evaluated `Util::get_local_path()` inside `includes/class-image-optimisation.php` when converting an image to both AVIF and WebP, preventing it from being unnecessarily called twice.

🎯 Why:
- Redundant regular expression compilation inside high-frequency loops (like HTML tag processing) adds unnecessary computational overhead.
- Parsing URLs and resolving local file paths via `Util::get_local_path()` is an expensive string operation. Calling it twice for the exact same source image URL when generating AVIF and WebP variants is wasteful.

📊 Impact:
- Reduces string operations and regex compilations per page load significantly. For a page with dozens of images and `srcset` variants, `preg_quote()` was previously executed hundreds of times redundantly.
- Saves at least one redundant `Util::get_local_path()` call per image processing attempt when dual format conversions are enabled.

🔬 Measurement:
- Inspect `includes/class-cache.php` and verify `$site_url_regex` is computed once.
- Check `includes/class-image-optimisation.php` and confirm `$source_image_path` is instantiated only when required and reused for both conversion format tests.

---
*PR created automatically by Jules for task [6273755455358734687](https://jules.google.com/task/6273755455358734687) started by @nilesh32236*

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved efficiency when rewriting CDN URLs.
  * Reduced repeated image path resolution during next-generation image processing.
  * Added guidance to avoid redundant operations in high-frequency parsing workflows.

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -51,3 +51,7 @@
 ## 2026-07-21 - Throttle Environment Failure Checks
 **Learning:** When performing expensive environment checks (like file I/O to check or write to `wp-config.php`) inside high-frequency hooks like `admin_init`, failing to throttle the execution when the check *fails* will cause the application to repeatedly retry the failing operation on every single page load.
 **Action:** Always unconditionally set the throttle transient (using `set_transient`) outside of success/failure conditional blocks to prevent constant retry loops.
+
+## 2025-01-22 - Redundant Operations in Parsers
+**Learning:** Functions called within high-frequency loops, such as parsing HTML tags, can cause significant performance degradation. Generating regular expressions via `preg_quote` and doing complex URL processing (like `Util::get_local_path`) for each tag or property evaluated can add substantial overhead across a large document.
+**Action:** Lift static computations and string transformations out of loops when parsing. Delay expensive lookups (like resolving local filesystem paths) with lazy evaluation, ensuring they're executed at most once per distinct resource.
--- a/includes/class-cache.php
+++ b/includes/class-cache.php
@@ -416,7 +416,9 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Cache' ) ) {
 				return $buffer;
 			}
 
-			$tags = new \WP_HTML_Tag_Processor( $buffer );
+			// Cache the expensive regex generation outside the loop to improve performance.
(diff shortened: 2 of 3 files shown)


nilesh32236 (commented):
## PR Review Summary

This PR makes two targeted performance improvements: extracting `preg_quote($site_url, '#')` out of a `WP_HTML_Tag_Processor` loop in `class-cache.php` (saving hundreds of redundant regex compilations per page), and lazily evaluating `Util::get_local_path()` in `class-image-optimisation.php` so it's computed at most once per image when both AVIF and WebP conversions are enabled. The changes are minimal, correct, and align exactly with the PR description and the documented learning in `.jules/bolt.md`.

**Ready to merge?** true

**Reasoning:** The implementation matches the PR description precisely with no bugs, security issues, dead code, or architecture concerns. Both changes are correct, well-scoped, and follow recommended patterns from the codebase's performance knowledge base. The edge cases (empty paths, null state, webp-only mode) are all handled correctly. This is ready to merge.

### Strengths

- **includes/class-cache.php:420** — Clean extraction: `preg_quote()` with the `#` delimiter is called once before the loop and stored in `$site_url_regex`, then reused in both the attribute loop (line 434) and srcset loop (line 451). This eliminates hundreds of redundant regex compilations per page without changing any behavior.
- **includes/class-image-optimisation.php:328** — Good lazy-evaluation pattern: `$source_image_path` is initialized to `null`, always computed in the AVIF block when needed, and conditionally recomputed in the WebP block only if still `null`. This correctly handles all three conversion modes (`avif`, `webp`, `both`) — in `both` mode the path is computed once and reused, in single-format modes it's computed only in the relevant block.
- **.jules/bolt.md:55** — The new learning entry documents the exact pattern applied in this PR, creating a useful knowledge base entry that future Bolt PRs can reference for similar redundant-computation optimizations in parsers.

PR #352, opened 2026-07-25
🚔 Warden: [Code Quality Improvement]

nilesh32236:
### What was cleaned up
Extracted the early logic checking for `removeCssJsHandle` into an early return guard clause and removed the redundant `! empty()` condition wrapping the `foreach` loops for `$remove_css_js_handle`.

### Why it was messy
The `remove_woocommerce_scripts` method was nested too deeply for conditions that could be resolved early (especially if the array to check is empty). Also, `! empty( $remove_css_js_handle )` wrapped around `foreach ( $remove_css_js_handle as $handle )` is redundant because PHP's `foreach` gracefully skips empty arrays natively, leading to unnecessary block nesting and visual clutter.

### How the new structure improves readability
Adding an early return at the top makes the baseline flow clearer, indicating exactly what conditions stop the rest of the script from processing. The removal of the `! empty( ... )` block flattens the execution path of the `foreach` loop, adhering closer to standard PHP and WPCS structural cleanliness by reducing cognitive complexity and horizontal indentation.

---
*PR created automatically by Jules for task [2595159310157211219](https://jules.google.com/task/2595159310157211219) started by @nilesh32236*

### DIFF
--- a/includes/class-main.php
+++ b/includes/class-main.php
@@ -665,10 +665,16 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Main' ) ) {
 		 * @since 1.0.0
 		 */
 		public function remove_woocommerce_scripts() {
-			$exclude_url_to_keep_js_css = array();
-			if ( isset( $this->options['file_optimisation']['excludeUrlToKeepJSCSS'] ) && ! empty( $this->options['file_optimisation']['excludeUrlToKeepJSCSS'] ) ) {
-				$exclude_url_to_keep_js_css = Util::process_urls( $this->options['file_optimisation']['excludeUrlToKeepJSCSS'] );
+			if ( empty( $this->options['file_optimisation']['removeCssJsHandle'] ) ) {
+				return;
+			}
 
+			$exclude_url_to_keep_js_css = array();
+			if ( ! empty( $this->options['file_optimisation']['excludeUrlToKeepJSCSS'] ) ) {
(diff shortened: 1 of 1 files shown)


nilesh32236 (commented):
## PR Review Summary

This PR refactors `remove_woocommerce_scripts()` in `includes/class-main.php` with an early return guard for `removeCssJsHandle`, removes redundant `isset()` calls (handled by `empty()`), and flattens the `foreach` loop by dropping the unnecessary `! empty()` wrapper. The changes are semantically correct, well-justified, and improve readability without introducing regressions.

**Ready to merge?** true

**Reasoning:** All changes are safe, semantically equivalent to the original, and strictly improve code clarity. No bugs, security issues, or regressions detected.

### Strengths

- **includes/class-main.php:668** — Early return guard cleanly separates the "nothing to do" case from the main logic, making the control flow immediately obvious.
- **includes/class-main.php:673** — Removing the redundant `isset()` (already covered by `empty()`) eliminates unnecessary boolean conjunction without changing behavior.
- **includes/class-main.php:677** — Extracting the exclusion-check block into its own `if (! empty(...))` after the data-loading `if` correctly separates "load data" from "act on data."
- **includes/class-main.php:704** — Removing the `! empty()` guard around `foreach` is idiomatic — PHP's `foreach` naturally handles empty arrays, and the early return already guarantees `removeCssJsHandle` is truthy.

PR #356, opened 2026-07-26
⚡ Bolt: Cache excludeDelayJS arrays to avoid parsing on every script tag

nilesh32236:
💡 What: Lifted `Util::process_urls` out of the `safe_minify_js` callback in `class-html.php` by caching it in a class property (`$exclude_delay_js`) during object instantiation.
🎯 Why: `safe_minify_js` is called via `preg_replace_callback` for *every single script tag* on the page. Processing the exclusion URLs (which involves `explode`, `array_map`, `array_unique`, etc.) dynamically on every single iteration creates a massive, hidden CPU overhead, significantly impacting Time to First Byte (TTFB). 
📊 Impact: Eliminates redundant array processing and string manipulation logic across the entire document parsing loop.
🔬 Measurement: Code syntax verified via `php -l`. Linter warnings fixed via `phpcbf`. Front-end built with `npm run build`. 

---
*PR created automatically by Jules for task [14864065076174802594](https://jules.google.com/task/14864065076174802594) started by @nilesh32236*

## Summary by CodeRabbit

- **Performance**
  - Improved JavaScript optimization efficiency by caching exclusion settings instead of recalculating them repeatedly.
  - Preserved existing delayed-minification exclusions while reducing processing overhead during page rendering.

- **Documentation**
  - Added guidance recommending that static configuration values be prepared once rather than recalculated inside frequently executed processing loops.

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -55,3 +55,7 @@
 ## 2025-01-22 - Redundant Operations in Parsers
 **Learning:** Functions called within high-frequency loops, such as parsing HTML tags, can cause significant performance degradation. Generating regular expressions via `preg_quote` and doing complex URL processing (like `Util::get_local_path`) for each tag or property evaluated can add substantial overhead across a large document.
 **Action:** Lift static computations and string transformations out of loops when parsing. Delay expensive lookups (like resolving local filesystem paths) with lazy evaluation, ensuring they're executed at most once per distinct resource.
+
+## 2025-01-22 - Array Processing in Regex Callback Loops
+**Learning:** Calling array processing operations (like `Util::process_urls`, which filters, maps and unique-ifies arrays) inside high-frequency regex callbacks (like parsing every `<script>` tag on a page for minification or delayJS) can severely impact performance.
+**Action:** When a static or configuration-based array needs to be processed to be used as exclusions or matches inside a regex callback loop, process and cache it once as a class property during instantiation instead of computing it dynamically for each match.
--- a/includes/minify/class-html.php
+++ b/includes/minify/class-html.php
@@ -57,6 +57,14 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Minify\HTML' ) ) {
 		private array $options;
 
 		/**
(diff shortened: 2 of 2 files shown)


nilesh32236 (commented):
## Executive Summary

**Purpose:** Lift Util::process_urls from the preg_replace_callback loop in safe_minify_js to the constructor, caching the result in a class property to avoid redundant array processing on every script tag.

**Risk:** 🟢 LOW — The change is isolated to one file, purely additive (class property + constructor cache), and removes redundant computation from a hot loop. No new functionality or API surface is introduced.

---

## PR Review Summary

The PR achieves exactly what it describes: caching the delayJS exclusion array in $this->exclude_delay_js during construction instead of recomputing Util::process_urls inside the preg_replace_callback loop. The implementation is correct, well-typed, and properly guarded with null coalescing. The bolt.md learning entry accurately documents the pattern for future reference.

**Ready to merge?** true

**Reasoning:** The implementation matches the PR description, is correct, introduces no new bugs, and meaningfully reduces CPU overhead in the HTML minification hot path.

### Strengths

- **includes/minify/class-html.php:65** — New typed private property $exclude_delay_js with proper PHPDoc block, correct visibility, and array type hint.
- **includes/minify/class-html.php:79** — Constructor caching correctly uses array_merge to combine hardcoded exclusion markers with processed user-configured exclusions, all computed once per request instead of per script tag.
- **includes/minify/class-html.php:83** — Defensive array sanitization (array_filter with strlen + array_values) strips empties and re-indexes, preventing edge-case issues in the foreach loop.
- **includes/minify/class-html.php:154** — Util class is loaded before HTML minify class in class-main.php includes(), guaranteeing process_urls is available at construct time.
- **:59** — Well-documented learning entry matching the actual performance fix, providing clear context and action for future contributors.

### Issues

- **MINOR:** includes/minify/class-html.php:341 — Redundant trim() calls inside the foreach loop — values are already trimmed by Util::process_urls during construction. Pre-existing behavior, not introduced by this PR, but still an unnecessary string operation per script tag.
  > Remove the trim() wrappers since the exclusions array is already trimmed.

PR #363, opened 2026-07-27
⚡ Bolt: [performance improvement] Cache expensive wp_upload_dir and home_url in static variables

nilesh32236:
💡 What: Implemented static caching for `wp_upload_dir()['basedir']` in `Img_Converter::add_img_into_queue()` and `home_url()` in `Util::get_local_path()`.

🎯 Why: These utility functions are called repeatedly inside `preg_replace_callback` loops (like `minify_css`, `add_delay_load_img`, and `replace_image_with_next_gen`) for every image or asset tag on a page. `home_url()` and `wp_upload_dir()` are expensive WordPress core functions that fire multiple filters and check transients/options. Statically caching them avoids massive performance bottlenecks.

📊 Impact: Reduces redundant executions of `home_url()` and `wp_upload_dir()` from O(N) (where N is the number of assets/images) to O(1) per request, significantly speeding up regex processing during page caching and minification.

🔬 Measurement: By profiling page load times and execution counts of `wp_upload_dir` and `home_url` with Xdebug or Query Monitor, one will observe these functions are only executed once per script lifecycle during minification.

---
*PR created automatically by Jules for task [18173089347690572347](https://jules.google.com/task/18173089347690572347) started by @nilesh32236*

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved performance when processing image queues and resolving local paths, especially during repeated operations.
  * Reduced redundant WordPress URL and upload-directory lookups within a request.

* **Documentation**
  * Added guidance for caching frequently used WordPress utility values during request processing.

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -36,8 +36,11 @@
 
 **Learning:** Calculating file sizes or counts via recursive directory scanning (`dirlist`) on every WP-admin load creates a severe bottleneck that can crash or slow down the settings page.
 **Action:** Always use WordPress Transients to cache expensive file system results, and invalidate these transients within `clear_cache` methods.
-## 2025-01-20 - Early Return Before Expensive Computations\n\n**Learning:** High-frequency operations like `Util::get_local_path` (which parses URLs) were being computed before short-circuit logic (`is_user_logged_in`, `empty( $href )`) in `minify_css` and `minify_js`. This caused significant performance overhead as they run on every page load.\n**Action:** Always place lightweight conditionals before heavy variable assignments. Avoid executing computationally expensive operations unless strictly necessary.
 
+## 2025-01-20 - Early Return Before Expensive Computations
+
+**Learning:** High-frequency operations like `Util::get_local_path` (which parses URLs) were being computed before short-circuit logic (`is_user_logged_in`, `empty( $href )`) in `minify_css` and `minify_js`. This caused significant performance overhead as they run on every page load.
+**Action:** Always place lightweight conditionals before heavy variable assignments. Avoid executing computationally expensive operations unless strictly necessary.
 
 ## 2025-01-22 - Bypassing WP_Filesystem for large file reads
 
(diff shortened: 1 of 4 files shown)


nilesh32236 (commented):
## Executive Summary

**Purpose:** Static caches expensive wp_upload_dir() and home_url() results inside loop-adjacent utility methods to avoid redundant WP core function calls (filters, options lookups) during regex-based page processing.

**Risk:** 🟢 LOW — Both changes are well-understood, safely guarded by a null-sentinel pattern, and maintain exact existing behavior. The cached values (upload basedir, home URL path) do not change mid-request in any standard WordPress scenario.

---

## PR Review Summary

Clean, minimal, and correct performance optimization. Both `get_local_path()` and `add_img_into_queue()` are called inside `preg_replace_callback` loops; caching `home_url()` and `wp_upload_dir()` results with function-local `static` variables eliminates redundant filter/option lookups per asset. The implementation exactly matches the PR description and follows the established patterns documented in `.jules/bolt.md`.

**Ready to merge?** true

**Reasoning:** Both changes are correct, safe, and align with the stated goals. No bugs, security concerns, or logic errors. The only consideration is a multisite/switch_to_blog edge case that is highly unlikely given the plugin's usage patterns.

### Strengths

- **includes/class-util.php:108** — Correctly uses `null` as the sentinel value instead of a falsy check like `empty()` or `!isset()`. This avoids subtle bugs where the cached value could legitimately be an empty string (e.g., `home_url()` with no path component), ensuring `wp_upload_dir()` is always evaluated at most once per request.
- **includes/class-img-converter.php:706** — Preserves the exact same trailing-slash normalization logic (`rtrim(... '/') . '/'`) from the original code without introducing any behavioral change. The static caching is a transparent performance improvement with zero side effects.

### Issues

- **MINOR:** `includes/class-util.php:108` — In a multisite network, if `get_local_path()` is called after `switch_to_blog()`, the `static $home_path` will reflect the original site's home URL, not the switched-to site's. This is because `home_url()` is context-aware (uses `get_current_blog_id()`), but the function-local `static` variable persists across blog switches for the request lifetime.
  > 💡 **How to fix:** Document this limitation in the method's docblock, or use a class-level static array keyed by `get_current_blog_id()` to support multisite context switches. However, since none of the current callers invoke `switch_to_blog()`, this is a theoretical edge case and not a blocking issue.

PR #367, opened 2026-07-28
⚡ Bolt: [Performance Improvement] Cache expensive WP core functions in loops

nilesh32236:
💡 What:
Implemented static caching for expensive WordPress core function calls (`home_url()`, `content_url()`, `wp_upload_dir()`, and `add_query_arg()`) that are repeatedly evaluated within parsing and normalization loops across `includes/class-util.php`, `includes/class-image-optimisation.php`, and `includes/class-img-converter.php`. The static arrays are explicitly keyed by `get_current_blog_id()` to ensure complete safety and accuracy in WordPress Multisite environments where the active blog context may switch during a single request.

🎯 Why:
Within high-frequency execution paths (such as `preg_replace_callback` iterating over all `img` tags or resolving hundreds of URLs on a page), constantly calling core functions like `home_url()` and `wp_upload_dir()` triggers numerous internal WordPress hooks, database lookups (for options), and string manipulations. This introduces significant, compounded latency to the page generation time. Caching these values reduces the work to simple, fast PHP array lookups after the initial evaluation.

📊 Impact:
Significantly reduces server-side processing time and memory overhead when serving un-cached pages with numerous assets or images. Eliminates hundreds of redundant internal WP core function executions per page load.

🔬 Measurement:
Deploy this change and profile the server response time (TTFB) or utilize Query Monitor to inspect function call times on a media-heavy front-end page with Image Optimization or Minification features enabled. You should observe a measurable decrease in execution time within `Util::get_local_path`, `normalize_url`, and the regex replacement callbacks.

---
*PR created automatically by Jules for task [4390406931692209432](https://jules.google.com/task/4390406931692209432) started by @nilesh32236*

## Summary by CodeRabbit

- **Performance**
  - Improved image URL and path processing by reusing repeated site configuration lookups during high-volume operations.

- **Multisite Compatibility**
  - Improved handling of URLs, uploads, converted images, and local paths across multiple WordPress sites and blogs.

- **Bug Fixes**
  - Increased consistency when resolving protocol-relative and relative image paths, while preserving existing local path behavior.

- **Documentation**
  - Added guidance on reducing repeated WordPress URL and path calculations in performance-sensitive loops.

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -59,3 +59,7 @@
 ## 2025-01-22 - Array Processing in Regex Callback Loops
 **Learning:** Calling array processing operations (like `Util::process_urls`, which filters, maps and unique-ifies arrays) inside high-frequency regex callbacks (like parsing every `<script>` tag on a page for minification or delayJS) can severely impact performance.
 **Action:** When a static or configuration-based array needs to be processed to be used as exclusions or matches inside a regex callback loop, process and cache it once as a class property during instantiation instead of computing it dynamically for each match.
+
+## 2025-01-28 - Caching WP Core Functions in Loops
+**Learning:** Executing WordPress core functions like `home_url()`, `content_url()`, or `wp_upload_dir()` repeatedly inside high-frequency loops (e.g., regex callbacks parsing hundreds of images or CSS tags) creates significant performance overhead due to redundant hook executions and string processing within WP Core.
+**Action:** When WordPress core utility functions that return static paths/URLs are needed inside parsing loops, cache their results in PHP `static` variables so they are computed only once per request and reused across all subsequent loop iterations.
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -36,8 +36,11 @@
 
 **Learning:** Calculating file sizes or counts via recursive directory scanning (`dirlist`) on every WP-admin load creates a severe bottleneck that can crash or slow down the settings page.
 **Action:** Always use WordPress Transients to cache expensive file system results, and invalidate these transients within `clear_cache` methods.
-## 2025-01-20 - Early Return Before Expensive Computations\n\n**Learning:** High-frequency operations like `Util::get_local_path` (which parses URLs) were being computed before short-circuit logic (`is_user_logged_in`, `empty( $href )`) in `minify_css` and `minify_js`. This caused significant performance overhead as they run on every page load.\n**Action:** Always place lightweight conditionals before heavy variable assignments. Avoid executing computationally expensive operations unless strictly necessary.
 
+## 2025-01-20 - Early Return Before Expensive Computations
+
+**Learning:** High-frequency operations like `Util::get_local_path` (which parses URLs) were being computed before short-circuit logic (`is_user_logged_in`, `empty( $href )`) in `minify_css` and `minify_js`. This caused significant performance overhead as they run on every page load.
+**Action:** Always place lightweight conditionals before heavy variable assignments. Avoid executing computationally expensive operations unless strictly necessary.
 
 ## 2025-01-22 - Bypassing WP_Filesystem for large file reads
 
@@ -61,5 +64,9 @@
 **Action:** When a static or configuration-based array needs to be processed to be used as exclusions or matches inside a regex callback loop, process and cache it once as a class property during instantiation instead of computing it dynamically for each match.
 
 ## 2025-01-28 - Caching WP Core Functions in Loops
+
 **Learning:** Executing WordPress core functions like `home_url()`, `content_url()`, or `wp_upload_dir()` repeatedly inside high-frequency loops (e.g., regex callbacks parsing hundreds of images or CSS tags) creates significant performance overhead due to redundant hook executions and string processing within WP Core.
+
 **Action:** When WordPress core utility functions that return static paths/URLs are needed inside parsing loops, cache their results in PHP `static` variables so they are computed only once per request and reused across all subsequent loop iterations.
+
+**Exception (multisite):** In WordPress multisite, contexts can switch (e.g., via `switch_to_blog()`). When caching WP core function results, key the static cache by `get_current_blog_id()` to ensure lookups resolve correctly when the active site context changes within a single request.
--- a/includes/class-image-optimisation.php
+++ b/includes/class-image-optimisation.php
@@ -387,6 +387,9 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Image_Optimisation' ) ) {
 				return $url;
 			}
 
+			// Cache home_url() once per blog for the request.
+			static $home_base = array();
+
 			// Protocol-relative URLs (e.g., //example.com/image.jpg).
 			if ( strpos( $url, '//' ) === 0 ) {
 				static $scheme = array();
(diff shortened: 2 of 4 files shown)


nilesh32236 (commented):
## Executive Summary

**Purpose:** Cache expensive WP core function calls (home_url, content_url, wp_upload_dir, add_query_arg) in high-frequency loops across two files using PHP static variables keyed by blog_id.

**Risk:** 🟢 LOW — Changes are confined to local static variable caches with no structural refactoring, no new dependencies, and proper multisite isolation via get_current_blog_id(). All original function call semantics are preserved. Three minor issues remain — all are uncached call sites in the same hot paths, not regressions.

---

## PR Review Summary

The PR correctly identifies hot-path functions and introduces static caching with multisite-aware blog_id keys. Implementation is clean, minimal, and well-aligned with the description. Three minor issues remain: two uncached home_url() call sites in normalize_url() within class-image-optimisation.php, and an uncached wp_normalize_path(ABSPATH) in class-img-converter.php's add_img_into_queue().

**Ready to merge?** true

**Reasoning:** All three remaining issues are minor (inconsistent caching, not bugs or regressions) and do not block merge. The core optimization logic is sound and well-tested by existing patterns.

### Strengths

- **includes/class-img-converter.php:460** — Correct use of separate static $home_url and $content_url arrays prevents accidental key collision and keeps cache granularity aligned with the cached value.
- **includes/class-image-optimisation.php:392** — The scheme extraction optimization is well-placed: home_url() inside wp_parse_url() is the most expensive part of the protocol-relative branch, and caching it eliminates redundant hook execution on every protocol-relative URL in the document.
- **includes/class-img-converter.php:526** — The get_img_url cache properly isolates home_url() lookups per blog_id while keeping the method signature unchanged — no API surface impact.

### Issues

- **MINOR:** `includes/class-image-optimisation.php:406` — home_url($url) on the root-relative path branch (line 406) and home_url($absolute_path) on the true-relative path branch (line 422) in normalize_url() are not cached, despite being called inside the img-tag while loop (line 189) and srcset foreach (line 207). These are the only two uncached home_url() call sites remaining in the normalize_url hot path.
  > 💡 **How to fix:** Cache the plain home_url() result in a blog-keyed static variable and reuse it at both call sites: use rtrim($home_base, '/') . '/' . ltrim($path, '/') instead of calling home_url() again. Alternatively, extend the existing $scheme static cache to include a 'base' key with the full home_url() value.
- **MINOR:** `includes/class-img-converter.php:738` — wp_normalize_path(ABSPATH) is called on every add_img_into_queue() invocation that passes the upload-dir check. add_img_into_queue is invoked from loop-heavy contexts (CSS minifier regex callback, attachment metadata size iteration). This is inconsistent with the caching approach applied to other static values in the same method.
  > 💡 **How to fix:** Cache wp_normalize_path(ABSPATH) in a blog-keyed static variable alongside $upload_dir, or hoist it to a static class property if ABSPATH is guaranteed constant across sites in a multisite context.

nilesh32236 (commented):
## Executive Summary

**Purpose:** Fix 6 performance/quality issues in JS frontend: replace setInterval+async with recursive setTimeout to prevent overlapping poll executions, migrate from Math.random() to useId(), cache DOM queries in focus trap, conditionally activate safety scan, and properly await suggestion fetching.

**Risk:** 🟢 LOW — All changes are isolated to JS frontend files using well-understood patterns (recursive setTimeout, React useId, ref caching). No PHP, security, or data-layer changes. One bounded concurrency concern exists but has a straightforward fix.

---

## PR Review Summary

Well-structured PR that correctly addresses real race conditions in async polling loops. The recursive setTimeout pattern in Dashboard.js and PageSpeedPanel.js prevents overlapping API calls that could occur with setInterval+async. The useId() migration, cached DOM queries, and conditional safety scan are solid quality improvements. One important issue found: in Dashboard.js, the recursive setTimeout can be overwritten by optimizeImages, creating a window for concurrent poll schedules to compound. Fix is straightforward. Two minor issues also noted for ConfirmDialog cleanup and lazyload activation guard.

**Ready to merge?** false

**Reasoning:** One important fix needed in Dashboard.js (pollingRef.current race window) where optimizeImages can fire mid-pollJobStatus, leading to concurrent poll schedules that compound. Fix is 3 lines and well-understood. Two minor improvements also recommended. All other changes are correct and merge-worthy.

### Strengths

- **src/components/Dashboard.js:151** — Correct recursive setTimeout pattern in pollJobStatus. Properly serializes async execution by calling setTimeout only after the previous poll resolves, preventing overlapping API calls that could occur with setInterval + async callbacks. The guard at line 209 correctly prevents rescheduling when pollingRef.current has been externally nullified.
- **src/components/PageSpeedPanel.js:147** — Consistent recursive setTimeout fix applied in pollForResults. The self-scheduling approach ensures each poll tick waits for the previous getPagespeedResults() call to complete before scheduling the next, preventing concurrent scan status requests.
- **src/components/common/Tooltip.js:19** — Clean migration from Math.random() to React 18's useId(). Provides collision-free, SSR-safe IDs and eliminates the useRef wrapper. The id value correctly connects aria-describedby to the tooltip span's id attribute.
- **src/components/common/ConfirmDialog.js:68** — Caching querySelectorAll results in focusableRef via useEffect avoids redundant DOM queries on every Tab keydown — a solid accessibility micro-optimization.
- **src/lazyload.js:348** — Well-factored extraction of startSafetyScan into a named helper with early-return guard. Conditional activation (only when MutationObserver detects lazy elements or initial elements exist) avoids running an unnecessary 10s interval when the DOM has no lazy content.
- **src/components/PerformanceAudit.js:292** — Moving fetchSuggestions from a fire-and-forget .then().catch() to an explicit await with inner try/catch eliminates unhandled promise rejection risk. The sequential dependency between scan completion and suggestions fetching is now explicit, and scan UI state changes remain cleanly separated from the suggestions fetch.

### Issues

- **IMPORTANT:** `src/components/Dashboard.js:209` — Race window: when optimizeImages fires during an in-flight pollJobStatus (setting pollingRef.current to a new setTimeout), the completing pollJobStatus overwrites pollingRef.current when rescheduling, losing the reference to optimizeImages's timeout. Both timeouts remain active, leading to concurrent pollJobStatus invocations that each reschedule themselves, compounding over cycles.
  > 💡 **How to fix:** Capture pollingRef.current at the start of pollJobStatus and only reschedule if it hasn't been externally changed. This ensures external callers (like optimizeImages) who set a fresh timeout take precedence over the stale scheduled tick from the completing poll execution.
- **MINOR:** `src/components/common/ConfirmDialog.js:68` — focusableRef.current is not cleared when the dialog closes (isOpen→false), retaining stale DOM element references until the dialog reopens.
  > 💡 **How to fix:** Reset focusableRef.current to an empty array in the effect cleanup when isOpen becomes false, so unmounted DOM nodes are released for garbage collection.
- **MINOR:** `src/lazyload.js:383` — startSafetyScan() is called unconditionally for every MutationObserver callback where nodeType===1, even if the added node contains no lazy-loadable elements.
  > 💡 **How to fix:** Guard the startSafetyScan() call behind a check for whether the node subtree actually contains lazy elements, e.g. `if (node.matches(LAZY_SELECTOR) || node.querySelector(LAZY_SELECTOR)) startSafetyScan()`. This avoids unnecessary function calls on non-lazy DOM mutations.

PR #479, opened 2026-07-31
⚡ Bolt: Cache content_url in asset minification

nilesh32236:
💡 **What:** 
Added a static PHP array keyed by `get_current_blog_id()` inside `minify_css` and `minify_js` in `includes/class-main.php` to cache the result of `content_url( 'cache/wppo/min/css/' )` and `content_url( 'cache/wppo/min/js/' )`.

🎯 **Why:** 
`content_url()` is a WordPress core function that executes hooks and string manipulation. `minify_css` and `minify_js` are executed for *every single* CSS link and JS script tag on a page. Repeatedly calling `content_url()` creates redundant overhead. By caching it statically, we execute it only once per site during the request lifecycle.

📊 **Impact:** 
Eliminates redundant WordPress hook executions and string processing when multiple CSS or JS files are enqueued on the frontend, saving minor CPU cycles across every frontend page load.

🔬 **Measurement:** 
Profile a page load with multiple un-minified local CSS and JS files using Query Monitor or Xdebug. You should observe `content_url()` being invoked significantly fewer times for the `wppo/min/` paths.

---
*PR created automatically by Jules for task [4889929146123090468](https://jules.google.com/task/4889929146123090468) started by @nilesh32236*

## Summary by CodeRabbit

* **Performance**
  * Improved CSS and JavaScript asset minification efficiency on multisite installations.
  * Preserved generated asset URLs, including minified filenames and modification-time parameters.


.jules/bolt.md
```diff
@@ -69,4 +69,4 @@
...
 
-**Exception (multisite):** In WordPress multisite, contexts can switch (e.g., via `switch_to_blog()`). When caching WP core function results, key the static cache by `get_current_blog_id()` to ensure lookups resolve correctly when the active site context changes within a single request.
\ No newline at end of file
+**Exception (multisite):** In WordPress multisite, contexts can switch (e.g., via `switch_to_blog()`). When caching WP core function results, key the static cache by `get_current_blog_id()` to ensure lookups resolve correctly when the active site context changes within a single request.## 2026-07-31 - Cache content_url in Asset Minification Loops\n\n**Learning:** Executing WordPress core functions like `content_url()` repeatedly inside high-frequency execution paths (like when rewriting URLs for CSS and JS tags during minification) can add redundant performance overhead because it executes WP hooks for every replaced asset.\n**Action:** Declare a PHP static array keyed by `get_current_blog_id()` to execute `content_url()` once per site per request and reuse the cached URL prefix during minification processes.
```
nilesh32236, line 72: 🔵 **MINOR**: The new learning entry '## 2026-07-31 - Cache content_url in Asset Minification Loops' was appended directly onto the last line of the file with literal \n escape sequences and no separating blank line, so the file now ends in a single 820-byte line of raw '\n' text. The markdown header and paragraphs are not real newlines, which breaks the document structure for the autonomous agent workflows that read this file.

> 💡 **How to fix:** Append the new entry on its own lines with a preceding blank line instead of embedding literal \n escapes.

```suggestion
**Exception (multisite):** In WordPress multisite, contexts can switch (e.g., via `switch_to_blog()`). When caching WP core function results, key the static cache by `get_current_blog_id()` to ensure lookups resolve correctly when the active site context changes within a single request.

## 2026-07-31 - Cache content_url in Asset Minification Loops

**Learning:** Executing WordPress core functions like `content_url()` repeatedly inside high-frequency execution paths (like when rewriting URLs for CSS and JS tags during minification) can add redundant performance overhead because it executes WP hooks for every replaced asset.

**Action:** Declare a PHP static array keyed by `get_current_blog_id()` to execute `content_url()` once per site per request and reuse the cached URL prefix during minification processes.

**Exception (filterable paths):** `content_url()` runs `apply_filters( 'content_url', $url, $path )`, so path-sensitive filters may return different URLs per path. When `has_filter( 'content_url' )` is not `false`, call `content_url()` with the complete asset path (including `basename( $cached_file )`) on every rewrite instead of reusing a cached prefix; only use the per-blog cached prefix when no filter is registered.
```

nilesh32236, line 72: 🔵 **MINOR**: Per the open CodeRabbit thread, the bolt.md guidance (lines 66-72) still instructs agents to cache WP core function results in static variables unconditionally and does not document the has_filter('content_url') path-sensitive exception that the PR's own code implements. Without this documented, future agents will reintroduce the bug by dropping the filter guard.

> 💡 **How to fix:** Document the caveat in bolt.md: when has_filter('content_url') is not false, require full-path content_url() calls per rewrite rather than reusing a cached prefix; retain per-blog caching only when no content_url filter is registered.

```suggestion
Document the caveat in bolt.md: when has_filter('content_url') is not false, require full-path content_url() calls per rewrite rather than reusing a cached prefix; retain per-blog caching only when no content_url filter is registered.
```

nilesh32236 (commented):
## Executive Summary

**Purpose:** Caches the result of content_url() for wppo/min/ asset prefixes in PHP static arrays keyed by get_current_blog_id() inside minify_css/minify_js, with a has_filter('content_url') guard that falls back to full-path calls when a filter could alter the URL.

**Risk:** 🟢 LOW — The change is a conservative micro-optimization confined to two frontend tag-rewrite paths. Behavior parity with the previous full-path content_url() call is preserved in both branches (verified against WP core content_url() semantics), the code passes WPCS lint and php -l, and the path-sensitive filter edge case is explicitly guarded.

---

## MR Review Summary

A correct, low-risk performance tweak: content_url() is now computed once per blog per request in the common no-filter case, and the has_filter('content_url') guard correctly reverts to full-path calls when any filter could rewrite the asset URL path. The remaining problems are documentation-only: the new entry in .jules/bolt.md is malformed (literal \n appended to the last line) and fails to record the filter caveat that the code itself implements.

**Ready to merge?** true

**Reasoning:** No critical or important defects: the static cache is correctly blog-scoped, behavior is identical to the previous code when no content_url filter exists, and path-sensitive filters are handled by the fallback branch. Only minor doc-formatting/maintainability cleanup remains, so the PR is safe to merge.

### Strengths

- **includes/class-main.php:1557** — The `false !== has_filter( 'content_url' )` guard is the right call: WP core's content_url() runs apply_filters('content_url', $url, $path), so a path-sensitive filter can return a different URL for 'cache/wppo/min/css/' than for the full 'cache/wppo/min/css/<file>.css'. Branching to full-path calls when any filter is registered (and has_filter returns an int priority, hence the !== false check) is exactly the fix the review threads requested and it is implemented correctly.
- **includes/class-main.php:1560** — Keying the static cache by get_current_blog_id() correctly handles multisite switch_to_blog() contexts, matching the existing transient-key isolation convention documented in AGENTS.md. In single-site installs this is a constant, so there is no measurable overhead.
- **includes/class-main.php:1566** — Behavior parity is preserved: with no filter registered, content_url('cache/wppo/min/css/') . basename() produces byte-identical output to the prior content_url('cache/wppo/min/css/'.basename()), so existing cached-file URLs (and the ?ver=filemtime query string) are unchanged.

### Issues

- 🔵 **MINOR:** `includes/class-main.php:1560` — The 12-line URL-resolution block (basename + has_filter guard + per-blog static cache) is duplicated verbatim between minify_css (lines 1556-1567) and minify_js (lines 1617-1628). This invites drift — e.g., a future fix to the css branch could be missed in the js branch. **[medium confidence]**
  > 💡 **How to fix:** Extract a small private helper, e.g. resolve_minified_url( string $dir, string $cached_file ): string, that returns the content_url prefix (cached per blog when no filter exists) concatenated with the basename, and call it from both minify_css and minify_js.
- 🔵 **MINOR:** `includes/class-main.php:1561` — get_current_blog_id() is invoked on every rewritten asset even when the static cache is already warm for that blog, slightly undermining the 'cheapest check first' pattern already used elsewhere in this method (see bolt.md short-circuit guidance). **[low confidence]**
  > 💡 **How to fix:** Guard with empty( $content_url_base ) before calling get_current_blog_id(), or simply check isset( $content_url_base[ $blog_id ] ) with a single look-up, so the per-iteration work is one array lookup instead of an extra function call. The overhead is negligible; this is purely a consistency nicety.

PR #498, opened 2026-08-01
[Autofix] [WP Monitor] Honor DONOTCACHEPAGE constant in cache pipeline

### LINKED ISSUE #487: [WP Monitor] Honor DONOTCACHEPAGE constant in cache pipeline
## Summary
Core, WooCommerce, and many plugins set the `DONOTCACHEPAGE` constant to signal "do not cache this page". The plugin's custom `is_not_cacheable()`/`maybe_store_cache()` logic does not check it, which can cause stale cached copies of pages the ecosystem explicitly opted out of caching.

## Current Implementation
- `includes/class-cache.php` — custom `is_not_cacheable()` (~line 627) / `maybe_store_cache()` (~line 780); no `DONOTCACHEPAGE` check.
- `includes/class-advanced-cache-handler.php` — filesystem drop-in serving has cart/cookie/role exclusions but no `DONOTCACHEPAGE` marker support.

## Recommended Change
- Add `defined( 'DONOTCACHEPAGE' ) && DONOTCACHEPAGE` as a skip condition in `is_not_cacheable()`.
- Mirror in the advanced-cache drop-in via a marker file or rewrite check (the drop-in never boots WordPress, so it cannot read the constant directly).

## Fallback Strategy
None needed — purely additive check.

## Risk
**Low** — isolated to two cache-decision functions.

## Targets
- `includes/class-cache.php` — `is_not_cacheable()` / `maybe_store_cache()`
- `includes/class-advanced-cache-handler.php` — drop-in serving logic

nilesh32236:
## What Was Changed

# Fix: Honor `DONOTCACHEPAGE` constant in the cache pipeline (#487)

### What Was Done

Core, WooCommerce, and third-party plugins signal "do not cache this page" via the `DONOTCACHEPAGE` constant. The plugin now honors that constant on both the write path (static HTML generation) and the serve path (the `advanced-cache.php` drop-in), so opted-out pages are never cached or served from stale static files.

### Approach

The cache decision is made in two independent places, so the fix targets both:

1. **Write path (PHP, runs with WordPress booted):** `Cache::is_not_cacheable()` and `Cache::maybe_store_cache()` now return early when `defined( 'DONOTCACHEPAGE' ) && DONOTCACHEPAGE`. This prevents a freshly rendered page from ever being written to the static cache.

2. **Serve path (drop-in, runs before WordPress boots):** The drop-in cannot read the PHP constant, so the plugin writes a filesystem marker — a hidden `.wppo-no-cache` file beside `index.html` — the first time a request hits a `DONOTCACHEPAGE` page. The generated `advanced-cache.php` now checks for that marker and, if present, returns early so WordPress boots and renders the dynamic page instead of serving a stale copy.

3. **Marker lifecycle:** Markers are purged on single-page cache clear and on post invalidation (self-healing — if a plugin stops setting the constant, caching resumes on the next request after the marker is removed). Full cache clears already delete the whole domain directory, so markers are removed automatically there.

4. **Upgrade path:** Existing installs keep their old drop-in (which lacks the marker check), so a one-time version-upgrade routine regenerates the drop-in and clears the full cache once on version change. This removes pre-existing stale `DONOTCACHEPAGE` pages that the old drop-in would otherwise keep serving forever.

### Key Changes

- **`includes/class-cache.php`**
  - `is_not_cacheable()`: returns `true` (not cacheable) when `DONOTCACHEPAGE` is defined and truthy.
  - `maybe_store_cache()`: returns `false` (don't store) when `DONOTCACHEPAGE` is defined and truthy — the final gate in `save_cache_files()`, covering both the WP 6.9+ buffer path and the legacy `template_redirect` path.
  - New `maybe_mark_page_not_cacheable()` helper: writes the `.wppo-no-cache` marker and purges any pre-existing static files (html, gz, and role variants) for the current URL. Guarded by a `no_cache_marker_written` property so it runs at most once per request.
  - `invalidate_dynamic_static_html()` and `clear_cache()` (single-page path): now also delete the `.wppo-no-cache` marker so caching can resume when the constant is no longer set.

- **`includes/class-advanced-cache-handler.php`**
  - `create()`: the generated `advanced-cache.php` now checks `file_exists( dirname( $file_path ) . '/.wppo-no-cache' )` and returns early when the marker is present, before any serve block runs. This is the serve-time mirror of the PHP-side check.

- **`includes/class-main.php`**
  - New `maybe_run_version_upgrade()` (hooked to `admin_init`): on a version change it regenerates the drop-in via `Advanced_Cache_Handler::create()` (which still respects foreign drop-ins) and clears the full cache once, then records `wppo_version`.

### Verification

- `npm run lint:js` — pass
- `composer lint` — pass
- `npm test` — 18 suites / 217 tests pass
- `npm run build` — compiled successfully (no JS/SCSS changed, so committed `build/` output is unchanged)

Note: `composer test` reports "No tests executed!" — this is pre-existing (PHPUnit 11 does not discover the repo's `class-cache-test.php`-style file names, which don't match the `*Test.php` suffix requirement); the repo has no runnable PHP unit tests per `AGENTS.md`.

## Files Changed

- `includes/class-advanced-cache-handler.php`
- `includes/class-cache.php`
- `includes/class-main.php`

## Test Plan

---

*🤖 Auto-generated by opencode-ai-reviewer from branch `autofix/issue-487`.*


nilesh32236 (commented):
## Executive Summary

**Purpose:** Honors the DONOTCACHEPAGE constant across the static-HTML cache pipeline: the write path (Cache::is_not_cacheable/maybe_store_cache) refuses to store opted-out pages and writes a .wppo-no-cache filesystem marker that the generated advanced-cache.php drop-in checks before serving, plus a one-time version-upgrade routine that regenerates the drop-in and clears the full cache.

**Risk:** 🔴 HIGH — The core three-file fix is well-constructed and double-gated on the write path, but the commit ships ~4,700 dev-only vendor files (~525K lines) into the tracked vendor/ tree in violation of the production-only vendor convention, the serve path cannot stop serving an already-cached page that later sets DONOTCACHEPAGE (WordPress never boots to write the marker), the marker never self-heals so transiently/conditionally flagged URLs permanently lose static caching, and the version-gated migration cannot fire without a WPPO_VERSION bump.

**Breaking Changes:**
- ⚠️ One-time full static cache clear and advanced-cache.php drop-in regeneration on first admin page load after a version change
- ⚠️ URLs that ever trigger DONOTCACHEPAGE get a permanent .wppo-no-cache marker until a cache clear or post invalidation removes it
- ⚠️ vendor/ now tracks dev-only dependencies instead of the production-only set

---

## MR Review Summary

The dual write-path/serve-path design is sound: the write path is gated at the final storage decision in save_cache_files(), the drop-in marker check is a clean mirror with consistent path derivation (homepage, trailing slash, role variants), and the at-most-once guard prevents redundant I/O. However, the PR must not merge as-is: it commits the entire dev-dependency vendor tree contrary to AGENTS.md, the marker never self-heals (permanently disabling caching for URLs that only transiently set the constant), already-cached pages transitioning to DONOTCACHEPAGE keep being served stale because the drop-in serves before WordPress boots, the marker write can fail silently when the cache directory is absent, and the upgrade migration is a no-op for existing 1.8.0 installs unless WPPO_VERSION is bumped.

**Ready to merge?** false

**Reasoning:** The write-path gating and the serve-path marker are correct for pages never cached or rendered after the constant is set, but the dev vendor tree (~4,700 files) violates the production-only vendor policy, the marker lifecycle does not actually self-heal, the marker write silently fails when the parent cache directory is absent, and the version-gated migration cannot fire unless WPPO_VERSION is bumped in the same release. Several of these require design-level decisions or manual process steps rather than mechanical fixes, so an automated agent should not merge as-is.

### Strengths

- **includes/class-cache.php:828** — Placing the DONOTCACHEPAGE gate in maybe_store_cache() as the final gate inside save_cache_files() correctly covers both the WP 6.9+ enhancement-buffer path and the legacy template_redirect path, catching constants defined mid-render after the earlier is_not_cacheable() checks already ran, while the 'html' === $type condition keeps CSS/JS minification unaffected.
- **includes/class-cache.php:641** — The no_cache_marker_written property guarantees the marker write and stale-file purge run at most once per request even though is_not_cacheable() is reachable from multiple call sites (combine_css, start_output_buffer, process_buffer_for_cache, stash_cache), avoiding duplicate filesystem work.
- **includes/class-advanced-cache-handler.php:155** — The generated drop-in marker check is placed before any serve logic and keys off the page directory, so it also covers role-variant files; the marker path (dirname of the cached file) exactly matches the write-path marker including homepage and trailing-slash normalization, and the emitted code is lint-clean (php -l passes).
- **includes/class-cache.php:899** — The marker lifecycle is integrated into the existing invalidation paths — single-page clear_cache(), invalidate_dynamic_static_html() smart purge, and the full recursive clear — so the marker's lifetime is consistent with every other cache artifact and caching can resume once the constant is no longer set.
- **includes/class-main.php:556** — The one-time version-upgrade routine is a sensible migration: it regenerates the drop-in via Advanced_Cache_Handler::create(), which still respects foreign drop-ins via foreign_dropin_present(), and clears the full cache once to purge stale pages the old drop-in would keep serving.

### Issues

- 🔵 **MINOR:** `includes/class-cache.php:640` — No automated tests cover the new marker lifecycle (write-once guard, purge on single-page clear and post invalidation, version-upgrade drop-in regeneration), and the repo has no runnable PHP unit tests per AGENTS.md, so the highest-risk write/serve path interaction is shipped verified only by lint and manual checks. **[medium confidence]**
  > 💡 **How to fix:** Add a PHP integration test or a documented manual test scenario covering: (1) a DONOTCACHEPAGE request writes the marker and purges existing html/gz/role variants, (2) clear_cache()/invalidate_dynamic_static_html() removes the marker, and (3) the generated drop-in returns early when the marker exists.

nilesh32236:
**Inline comment (vendor/composer/autoload_classmap.php:1)**

🔵 **MINOR**: The commit adds ~4,700 files (~525K lines) of require-dev dependencies to the tracked vendor/ directory (phpunit, mockery, hamcrest, brain/monkey, antecedent/patchwork, squizlabs/php_codesniffer, wp-coding-standards/wpcs, phpcompatibility, phpcsstandards, sebastian/*, nikic/php-parser, theseer/tokenizer, etc.) and regenerates the composer autoloader to reference them. AGENTS.md explicitly requires vendor/ to be production-only (run 'composer release' before committing), and the parent commit contained exactly the production packages (voku/html-min, matthiasmullie/minify, woocommerce/action-scheduler). The PR description only lists six vendor/composer files, understating the change, and .distignore excludes only a few of these paths, so the release ZIP would ship test/lint tooling.

> 💡 **How to fix:** Run 'composer release' (composer install --no-dev --optimize-autoloader --ignore-platform-reqs), verify vendor/ no longer contains phpunit/mockery/phpcs/wpcs/brain/antecedent, and recommit the production-only vendor/ plus the regenerated composer autoload metadata. Consider a CI guard that fails when vendor/ contains require-dev packages.

```suggestion
composer release
# then git add vendor/ includes/ and commit only production deps
```

includes/class-advanced-cache-handler.php
```diff
@@ -152,6 +152,11 @@ public static function create(): void {
 			'$file_path      = rtrim( $file_path, \'/\' );' . PHP_EOL .
 			'$gzip_file_path = $file_path . \'.gz\';' . PHP_EOL . PHP_EOL .
 
+			'$no_cache_marker = dirname( $file_path ) . \'/.wppo-no-cache\';' . PHP_EOL .
+			'if ( file_exists( $no_cache_marker ) ) {' . PHP_EOL .
```
nilesh32236, line 165: 🔵 **MINOR**: Residual stale-serve gap: the drop-in serves a cached index.html before WordPress boots, so for a page that is already cached when a plugin first sets DONOTCACHEPAGE, maybe_mark_page_not_cacheable() never runs and the marker is never written — the stale page keeps being served indefinitely until a cache clear, post save, or the one-time upgrade purge. This is exactly the issue #487 transition scenario, so the PR's claim that opted-out pages are 'never served from stale static files' is overstated. **[medium confidence]**

> 💡 **How to fix:** Document this as a best-effort limitation and scope the PR's guarantee to pages rendered at least once after the constant is set; optionally bound the stale window with a low-cost serve-path freshness check (e.g. skip serving when the cached file mtime is older than the option wppo_cache_last_cleared_time or a TTL).

```suggestion
Document this as a best-effort limitation and scope the PR's guarantee to pages rendered at least once after the constant is set; optionally bound the stale window with a low-cost serve-path freshness check (e.g. skip serving when the cached file mtime is older than the option wppo_cache_last_cleared_time or a TTL).
```

includes/class-cache.php
```diff
@@ -778,6 +825,11 @@ private function save_processed_buffer( string $buffer, string $file_path ): voi
 		 * @since 1.0.0
 		 */
 		private function maybe_store_cache() {
+			if ( defined( 'DONOTCACHEPAGE' ) && DONOTCACHEPAGE ) {
+				$this->maybe_mark_page_not_cacheable();
+				return false;
```
nilesh32236, line 1153: 🔵 **MINOR**: The marker never self-heals: it is only removed by post-save invalidation, a single-page clear, or a full clear. A URL that sets DONOTCACHEPAGE only transiently/conditionally (campaign windows, time-of-day, per-session, membership or A/B conditions not tied to post saves) keeps its marker permanently, so the drop-in boots WordPress and re-renders the page on every request even after the constant stops being set — a silent, permanent caching loss that contradicts the PR's 'self-healing' claim. **[medium confidence]**

> 💡 **How to fix:** Delete the marker on the successful cacheable write path so caching resumes automatically once the constant is no longer set; the next request that actually sets DONOTCACHEPAGE re-creates the marker and purges stale files, so behavior converges safely.

```suggestion
if ( 'html' === $type ) {
	$this->delete_cache_files( trailingslashit( dirname( $file_path ) ) . '.wppo-no-cache' );
}
```

includes/class-main.php
```diff
@@ -542,6 +543,30 @@ public function maybe_fix_wp_cache(): void {
...
+		 * Run one-time upgrade routines when the plugin version changes.
+		 *
+		 * Regenerates the advanced-cache.php drop-in so it honours the
+		 * DONOTCACHEPAGE no-cache marker, then clears the full cache once to
+		 * remove any pre-existing stale pages the old drop-in would keep serving.
+		 *
+		 * @return void
+		 * @since 1.9.0
+		 */
+		public function maybe_run_version_upgrade(): void {
```
nilesh32236, line 797: 🔵 **MINOR**: maybe_run_version_upgrade() is hooked to admin_init with no capability check. admin_init fires for any authenticated user who can load any admin page (including subscribers) or POST to admin-ajax.php, so an unprivileged user can trigger a full cache wipe (Cache::clear_cache()) and advanced-cache.php regeneration on every version bump. Every other admin-triggered operation in this plugin (REST endpoints, admin notices, metabox) is gated on manage_options.

> 💡 **How to fix:** Add a current_user_can('manage_options') guard as the first statement in maybe_run_version_upgrade() so only administrators can trigger the destructive upgrade routine.

```suggestion
public function maybe_run_version_upgrade(): void {
	if ( ! current_user_can( 'manage_options' ) ) {
		return;
	}

	$installed_version = get_option( 'wppo_version', '' );
```

includes/class-cache.php
```diff
@@ -617,6 +627,37 @@ private function minify_buffer( $buffer ) {
...
+		 * @return void
+		 * @since 1.9.0
+		 */
+		private function maybe_mark_page_not_cacheable(): void {
+			if ( $this->no_cache_marker_written ) {
+				return;
+			}
+			$this->no_cache_marker_written = true;
+
+			$html_file_path = $this->get_cache_file_path( 'html' );
```
nilesh32236, line 934: 🔵 **MINOR**: maybe_mark_page_not_cacheable() writes the marker with put_contents() without ensuring the parent directory exists (WP_Filesystem_Direct::put_contents() does not create parent directories) and ignores the return value. On a DONOTCACHEPAGE URL that has never been cached (or whose directory was removed), the marker write silently fails and the serve-path protection never engages; the static html/gz files may already have been purged, so a later request without the constant can re-cache the page with no marker present.

> 💡 **How to fix:** Create the cache directory before writing the marker (via $this->prepare_cache_dir()), skip the write when the marker already exists, and bail out/log when put_contents() returns false instead of silently no-op'ing.

```suggestion
$fs = $this->get_filesystem();
if ( ! $fs || $fs->exists( $marker_path ) ) {
	return;
}

if ( ! $this->prepare_cache_dir() ) {
	return;
}

$written = $fs->put_contents( $marker_path, (string) time(), FS_CHMOD_FILE );
if ( ! $written ) {
	return;
}
```

includes/class-main.php
```diff
@@ -542,6 +543,30 @@ public function maybe_fix_wp_cache(): void {
...
+			$installed_version = get_option( 'wppo_version', '' );
+
+			if ( version_compare( (string) $installed_version, WPPO_VERSION, '>=' ) ) {
```
nilesh32236, line 805: 🔵 **MINOR**: The version-upgrade migration cannot fire: WPPO_VERSION is still '1.8.0' (performance-optimisation.php:37) while the new members are documented @since 1.9.0, so existing 1.8.0 installs compare equal and never regenerate the marker-aware drop-in or run the one-time full clear — the serve-path half of the fix is a silent no-op for them until a future release bumps the version. Separately, wppo_version is never set on activation, so fresh installs redundantly run the full upgrade routine (drop-in regen + cache clear) on their first admin load. **[medium confidence]**

> 💡 **How to fix:** Bump WPPO_VERSION in the same release and align the @since tags with the shipped version; set the wppo_version option during activation (class-activate.php) so fresh installs skip the redundant upgrade routine.

```suggestion
define( 'WPPO_VERSION', '1.9.0' );
```

includes/class-cache.php
```diff
@@ -617,6 +627,37 @@ private function minify_buffer( $buffer ) {
...
+			$marker_path    = trailingslashit( dirname( $html_file_path ) ) . '.wppo-no-cache';
+
+			$fs = $this->get_filesystem();
+			if ( ! $fs ) {
+				return;
+			}
+
+			$fs->put_contents( $marker_path, (string) time(), FS_CHMOD_FILE );
```
nilesh32236: 🔵 **MINOR**: On every request to a page where DONOTCACHEPAGE is set, the marker is rewritten and the full stale-file purge (exists+delete checks plus a dirlist scan in delete_role_variant_files()) runs even when the marker already exists, adding unnecessary filesystem I/O to every dynamic render of high-traffic opted-out pages.

> 💡 **How to fix:** Early-return when the marker already exists so the write and purge run at most once on disk.

```suggestion
if ( $fs->exists( $marker_path ) ) {
	$this->no_cache_marker_written = true;
	return;
}
```

nilesh32236:
**Inline comment (includes/class-cache.php:779)**

🔵 **MINOR**: The guard `if ( ! $this->maybe_store_cache() && 'html' === $type )` evaluates maybe_store_cache() for every file type because the left operand runs first, so the DONOTCACHEPAGE branch — which writes the marker and purges html/role files — can fire on a 'css' save. Unreachable today because combine_css() is already gated by is_not_cacheable(), but the operand order is fragile and the side effects are surprising. **[medium confidence]**

> 💡 **How to fix:** Reorder the guard so maybe_store_cache() only runs for html writes.

```suggestion
if ( 'html' === $type && ! $this->maybe_store_cache() ) {
	return;
}
```

includes/class-cache.php
```diff
@@ -629,6 +670,12 @@ private function is_not_cacheable(): bool {
 				return true;
 			}
 
+			// Core, WooCommerce, and third-party plugins signal dynamic pages via DONOTCACHEPAGE.
+			if ( defined( 'DONOTCACHEPAGE' ) && DONOTCACHEPAGE ) {
+				$this->maybe_mark_page_not_cacheable();
```
nilesh32236, line 982: 🔵 **MINOR**: is_not_cacheable() is a predicate but now performs filesystem side effects (marker write + stale-file purge) via maybe_mark_page_not_cacheable(), and it is also used to gate CSS combining and buffer optimization, so the check/write coupling is surprising and hard to reason about despite the once-per-request guard. **[low confidence]**

> 💡 **How to fix:** Separate the pure cacheability check from the side-effecting marker write (or rename the helper to reflect that it both evaluates and records the no-cache decision) so callers that only need to test cacheability do not touch the filesystem.

```suggestion
Separate the pure cacheability check from the side-effecting marker write (or rename the helper to reflect that it both evaluates and records the no-cache decision) so callers that only need to test cacheability do not touch the filesystem.
```

includes/class-cache.php
```diff
@@ -629,6 +670,12 @@ private function is_not_cacheable(): bool {
 				return true;
 			}
 
+			// Core, WooCommerce, and third-party plugins signal dynamic pages via DONOTCACHEPAGE.
+			if ( defined( 'DONOTCACHEPAGE' ) && DONOTCACHEPAGE ) {
```
nilesh32236, line 981: 🔵 **MINOR**: Pages that set DONOTCACHEPAGE but were not previously excluded now also skip output-buffer optimizations (next-gen image conversion, HTML minification, CDN rewrite) because is_not_cacheable() gates process_buffer_for_cache(), start_output_buffer(), and combine_css(). This matches how other non-cacheable pages behave, but it is a behavior change worth confirming is intended. **[medium confidence]**

> 💡 **How to fix:** If buffer optimizations should still apply to DONOTCACHEPAGE pages, gate the DONOTCACHEPAGE early-return to the cache-storage decisions (maybe_store_cache/stash_cache) rather than the shared is_not_cacheable() optimization gate.

```suggestion
If buffer optimizations should still apply to DONOTCACHEPAGE pages, gate the DONOTCACHEPAGE early-return to the cache-storage decisions (maybe_store_cache/stash_cache) rather than the shared is_not_cacheable() optimization gate.
```

includes/class-main.php
```diff
@@ -542,6 +543,30 @@ public function maybe_fix_wp_cache(): void {
...
+				return;
+			}
+
+			Advanced_Cache_Handler::create();
+
+			Cache::clear_cache();
+
+			update_option( 'wppo_version', WPPO_VERSION, false );
```
nilesh32236, line 822: 🔵 **MINOR**: maybe_run_version_upgrade() records wppo_version even when Advanced_Cache_Handler::create() fails (e.g. transient WP_Filesystem failure), so the marker-aware drop-in regeneration is never retried and pre-existing stale DONOTCACHEPAGE pages keep being served. **[medium confidence]**

> 💡 **How to fix:** Only update wppo_version when the drop-in is actually regenerated (or when a foreign drop-in was detected and create() correctly returned early); otherwise leave the option unchanged so a later admin_init retries the upgrade.

```suggestion
Only update wppo_version when the drop-in is actually regenerated (or when a foreign drop-in was detected and create() correctly returned early); otherwise leave the option unchanged so a later admin_init retries the upgrade.
```

includes/class-main.php
```diff
@@ -328,6 +328,7 @@ private function includes(): void {
 		private function setup_hooks(): void {
 			add_action( 'admin_menu', array( $this, 'init_menu' ) );
 			add_action( 'admin_init', array( $this, 'maybe_fix_wp_cache' ) );
+			add_action( 'admin_init', array( $this, 'maybe_run_version_upgrade' ) );
```
nilesh32236, line 361: 🔵 **MINOR**: The upgrade routine is hooked to admin_init only, so sites updated via background auto-update or CLI (or whose admins do not visit wp-admin for a while) will not regenerate the marker-aware drop-in or clear stale DONOTCACHEPAGE pages until someone loads an admin page, silently delaying the fix for them. **[medium confidence]**

> 💡 **How to fix:** Also hook the one-time version check on 'upgrader_process_complete'/'activated_plugin', or run the version_compare on init since the option gate makes it one-time.

includes/class-cache.php
```diff
@@ -844,6 +896,7 @@ public function invalidate_dynamic_static_html( $page_id ): void {
 			$this->delete_role_variant_files( dirname( $html_file_path ) );
 			$this->delete_cache_files( $css_file_path );
 			$this->delete_cache_files( $used_css_path );
+			$this->delete_cache_files( trailingslashit( dirname( $html_file_path ) ) . '.wppo-no-cache' );
```
nilesh32236: 🔵 **MINOR**: invalidate_dynamic_static_html() removes the .wppo-no-cache marker only for the edited post; the smart-purge blocks for the home page and posts/archives/terms (lines 903-944) delete their cache files but not their markers, so a home/archive page that had a marker stays uncached even after its content changes. **[medium confidence]**

> 💡 **How to fix:** Delete the marker alongside the cache files in the home, blog, archive, and term purge blocks, reusing the same delete_cache_files(dirname-relative '.wppo-no-cache') pattern used for the post itself.

```suggestion
Delete the marker alongside the cache files in the home, blog, archive, and term purge blocks, reusing the same delete_cache_files(dirname-relative '.wppo-no-cache') pattern used for the post itself.
```

nilesh32236:
## 🤖 Autofix Review

### Summary

The dual write-path/serve-path design is sound: the write path is gated at the final storage decision in save_cache_files(), the drop-in marker check is a clean mirror with consistent path derivation (homepage, trailing slash, role variants), and the at-most-once guard prevents redundant I/O. However, the PR must not merge as-is: it commits the entire dev-dependency vendor tree contrary to AGENTS.md, the marker never self-heals (permanently disabling caching for URLs that only transiently set the constant), already-cached pages transitioning to DONOTCACHEPAGE keep being served stale because the drop-in serves before WordPress boots, the marker write can fail silently when the cache directory is absent, and the upgrade migration is a no-op for existing 1.8.0 installs unless WPPO_VERSION is bumped.

### Issues Found
- 🔵 **MINOR:** `vendor/composer/autoload_classmap.php:1` — The commit adds ~4,700 files (~525K lines) of require-dev dependencies to the tracked vendor/ directory (phpunit, mockery, hamcrest, brain/monkey, antecedent/patchwork, squizlabs/php_codesniffer, wp-coding-standards/wpcs, phpcompatibility, phpcsstandards, sebastian/*, nikic/php-parser, theseer/tokenizer, etc.) and regenerates the composer autoloader to reference them. AGENTS.md explicitly requires vendor/ to be production-only (run 'composer release' before committing), and the parent commit contained exactly the production packages (voku/html-min, matthiasmullie/minify, woocommerce/action-scheduler). The PR description only lists six vendor/composer files, understating the change, and .distignore excludes only a few of these paths, so the release ZIP would ship test/lint tooling.
  > 💡 **How to fix:** Run 'composer release' (composer install --no-dev --optimize-autoloader --ignore-platform-reqs), verify vendor/ no longer contains phpunit/mockery/phpcs/wpcs/brain/antecedent, and recommit the production-only vendor/ plus the regenerated composer autoload metadata. Consider a CI guard that fails when vendor/ contains require-dev packages.

```suggestion
composer release
# then git add vendor/ includes/ and commit only production deps
```
</details>
- 🔵 **MINOR:** `includes/class-advanced-cache-handler.php:156` — Residual stale-serve gap: the drop-in serves a cached index.html before WordPress boots, so for a page that is already cached when a plugin first sets DONOTCACHEPAGE, maybe_mark_page_not_cacheable() never runs and the marker is never written — the stale page keeps being served indefinitely until a cache clear, post save, or the one-time upgrade purge. This is exactly the issue #487 transition scenario, so the PR's claim that opted-out pages are 'never served from stale static files' is overstated. **[medium confidence]**
  > 💡 **How to fix:** Document this as a best-effort limitation and scope the PR's guarantee to pages rendered at least once after the constant is set; optionally bound the stale window with a low-cost serve-path freshness check (e.g. skip serving when the cached file mtime is older than the option wppo_cache_last_cleared_time or a TTL).
- 🔵 **MINOR:** `includes/class-cache.php:830` — The marker never self-heals: it is only removed by post-save invalidation, a single-page clear, or a full clear. A URL that sets DONOTCACHEPAGE only transiently/conditionally (campaign windows, time-of-day, per-session, membership or A/B conditions not tied to post saves) keeps its marker permanently, so the drop-in boots WordPress and re-renders the page on every request even after the constant stops being set — a silent, permanent caching loss that contradicts the PR's 'self-healing' claim. **[medium confidence]**
  > 💡 **How to fix:** Delete the marker on the successful cacheable write path so caching resumes automatically once the constant is no longer set; the next request that actually sets DONOTCACHEPAGE re-creates the marker and purges stale files, so behavior converges safely.

```suggestion
if ( 'html' === $type ) {
	$this->delete_cache_files( trailingslashit( dirname( $file_path ) ) . '.wppo-no-cache' );
}
```
</details>
- 🔵 **MINOR:** `includes/class-main.php:556` — maybe_run_version_upgrade() is hooked to admin_init with no capability check. admin_init fires for any authenticated user who can load any admin page (including subscribers) or POST to admin-ajax.php, so an unprivileged user can trigger a full cache wipe (Cache::clear_cache()) and advanced-cache.php regeneration on every version bump. Every other admin-triggered operation in this plugin (REST endpoints, admin notices, metabox) is gated on manage_options.
  > 💡 **How to fix:** Add a current_user_can('manage_options') guard as the first statement in maybe_run_version_upgrade() so only administrators can trigger the destructive upgrade routine.

```suggestion
public function maybe_run_version_upgrade(): void {
	if ( ! current_user_can( 'manage_options' ) ) {
		return;
	}

	$installed_version = get_option( 'wppo_version', '' );
```
</details>
- 🔵 **MINOR:** `includes/class-cache.php:646` — maybe_mark_page_not_cacheable() writes the marker with put_contents() without ensuring the parent directory exists (WP_Filesystem_Direct::put_contents() does not create parent directories) and ignores the return value. On a DONOTCACHEPAGE URL that has never been cached (or whose directory was removed), the marker write silently fails and the serve-path protection never engages; the static html/gz files may already have been purged, so a later request without the constant can re-cache the page with no marker present.
  > 💡 **How to fix:** Create the cache directory before writing the marker (via $this->prepare_cache_dir()), skip the write when the marker already exists, and bail out/log when put_contents() returns false instead of silently no-op'ing.

```suggestion
$fs = $this->get_filesystem();
if ( ! $fs || $fs->exists( $marker_path ) ) {
	return;
}

if ( ! $this->prepare_cache_dir() ) {
	return;
}

$written = $fs->put_contents( $marker_path, (string) time(), FS_CHMOD_FILE );
if ( ! $written ) {
	return;
}
```
</details>
- 🔵 **MINOR:** `includes/class-main.php:559` — The version-upgrade migration cannot fire: WPPO_VERSION is still '1.8.0' (performance-optimisation.php:37) while the new members are documented @since 1.9.0, so existing 1.8.0 installs compare equal and never regenerate the marker-aware drop-in or run the one-time full clear — the serve-path half of the fix is a silent no-op for them until a future release bumps the version. Separately, wppo_version is never set on activation, so fresh installs redundantly run the full upgrade routine (drop-in regen + cache clear) on their first admin load. **[medium confidence]**
  > 💡 **How to fix:** Bump WPPO_VERSION in the same release and align the @since tags with the shipped version; set the wppo_version option during activation (class-activate.php) so fresh installs skip the redundant upgrade routine.

```suggestion
define( 'WPPO_VERSION', '1.9.0' );
```
</details>
- 🔵 **MINOR:** `includes/class-cache.php:654` — On every request to a page where DONOTCACHEPAGE is set, the marker is rewritten and the full stale-file purge (exists+delete checks plus a dirlist scan in delete_role_variant_files()) runs even when the marker already exists, adding unnecessary filesystem I/O to every dynamic render of high-traffic opted-out pages.
  > 💡 **How to fix:** Early-return when the marker already exists so the write and purge run at most once on disk.

```suggestion
if ( $fs->exists( $marker_path ) ) {
	$this->no_cache_marker_written = true;
	return;
}
```
</details>
- 🔵 **MINOR:** `includes/class-cache.php:779` — The guard `if ( ! $this->maybe_store_cache() && 'html' === $type )` evaluates maybe_store_cache() for every file type because the left operand runs first, so the DONOTCACHEPAGE branch — which writes the marker and purges html/role files — can fire on a 'css' save. Unreachable today because combine_css() is already gated by is_not_cacheable(), but the operand order is fragile and the side effects are surprising. **[medium confidence]**
  > 💡 **How to fix:** Reorder the guard so maybe_store_cache() only runs for html writes.

```suggestion
if ( 'html' === $type && ! $this->maybe_store_cache() ) {
	return;
}
```
</details>
- 🔵 **MINOR:** `includes/class-cache.php:675` — is_not_cacheable() is a predicate but now performs filesystem side effects (marker write + stale-file purge) via maybe_mark_page_not_cacheable(), and it is also used to gate CSS combining and buffer optimization, so the check/write coupling is surprising and hard to reason about despite the once-per-request guard. **[low confidence]**
  > 💡 **How to fix:** Separate the pure cacheability check from the side-effecting marker write (or rename the helper to reflect that it both evaluates and records the no-cache decision) so callers that only need to test cacheability do not touch the filesystem.
- 🔵 **MINOR:** `includes/class-cache.php:674` — Pages that set DONOTCACHEPAGE but were not previously excluded now also skip output-buffer optimizations (next-gen image conversion, HTML minification, CDN rewrite) because is_not_cacheable() gates process_buffer_for_cache(), start_output_buffer(), and combine_css(). This matches how other non-cacheable pages behave, but it is a behavior change worth confirming is intended. **[medium confidence]**
  > 💡 **How to fix:** If buffer optimizations should still apply to DONOTCACHEPAGE pages, gate the DONOTCACHEPAGE early-return to the cache-storage decisions (maybe_store_cache/stash_cache) rather than the shared is_not_cacheable() optimization gate.
- 🔵 **MINOR:** `includes/class-main.php:567` — maybe_run_version_upgrade() records wppo_version even when Advanced_Cache_Handler::create() fails (e.g. transient WP_Filesystem failure), so the marker-aware drop-in regeneration is never retried and pre-existing stale DONOTCACHEPAGE pages keep being served. **[medium confidence]**
  > 💡 **How to fix:** Only update wppo_version when the drop-in is actually regenerated (or when a foreign drop-in was detected and create() correctly returned early); otherwise leave the option unchanged so a later admin_init retries the upgrade.
- 🔵 **MINOR:** `includes/class-main.php:331` — The upgrade routine is hooked to admin_init only, so sites updated via background auto-update or CLI (or whose admins do not visit wp-admin for a while) will not regenerate the marker-aware drop-in or clear stale DONOTCACHEPAGE pages until someone loads an admin page, silently delaying the fix for them. **[medium confidence]**
  > 💡 **How to fix:** Also hook the one-time version check on 'upgrader_process_complete'/'activated_plugin', or run the version_compare on init since the option gate makes it one-time.
- 🔵 **MINOR:** `includes/class-cache.php:899` — invalidate_dynamic_static_html() removes the .wppo-no-cache marker only for the edited post; the smart-purge blocks for the home page and posts/archives/terms (lines 903-944) delete their cache files but not their markers, so a home/archive page that had a marker stays uncached even after its content changes. **[medium confidence]**
  > 💡 **How to fix:** Delete the marker alongside the cache files in the home, blog, archive, and term purge blocks, reusing the same delete_cache_files(dirname-relative '.wppo-no-cache') pattern used for the post itself.
- 🔵 **MINOR:** `includes/class-cache.php:640` — No automated tests cover the new marker lifecycle (write-once guard, purge on single-page clear and post invalidation, version-upgrade drop-in regeneration), and the repo has no runnable PHP unit tests per AGENTS.md, so the highest-risk write/serve path interaction is shipped verified only by lint and manual checks. **[medium confidence]**
  > 💡 **How to fix:** Add a PHP integration test or a documented manual test scenario covering: (1) a DONOTCACHEPAGE request writes the marker and purges existing html/gz/role variants, (2) clear_cache()/invalidate_dynamic_static_html() removes the marker, and (3) the generated drop-in returns early when the marker exists.

### Strengths
- ✅ **includes/class-cache.php:828** — Placing the DONOTCACHEPAGE gate in maybe_store_cache() as the final gate inside save_cache_files() correctly covers both the WP 6.9+ enhancement-buffer path and the legacy template_redirect path, catching constants defined mid-render after the earlier is_not_cacheable() checks already ran, while the 'html' === $type condition keeps CSS/JS minification unaffected.
- ✅ **includes/class-cache.php:641** — The no_cache_marker_written property guarantees the marker write and stale-file purge run at most once per request even though is_not_cacheable() is reachable from multiple call sites (combine_css, start_output_buffer, process_buffer_for_cache, stash_cache), avoiding duplicate filesystem work.
- ✅ **includes/class-advanced-cache-handler.php:155** — The generated drop-in marker check is placed before any serve logic and keys off the page directory, so it also covers role-variant files; the marker path (dirname of the cached file) exactly matches the write-path marker including homepage and trailing-slash normalization, and the emitted code is lint-clean (php -l passes).
- ✅ **includes/class-cache.php:899** — The marker lifecycle is integrated into the existing invalidation paths — single-page clear_cache(), invalidate_dynamic_static_html() smart purge, and the full recursive clear — so the marker's lifetime is consistent with every other cache artifact and caching can resume once the constant is no longer set.
- ✅ **includes/class-main.php:556** — The one-time version-upgrade routine is a sensible migration: it regenerates the drop-in via Advanced_Cache_Handler::create(), which still respects foreign drop-ins via foreign_dropin_present(), and clears the full cache once to purge stale pages the old drop-in would keep serving.

### Iteration History

nilesh32236:

PR #537, opened 2026-08-04
⚡ Bolt: Cache content_url in minifiers

nilesh32236:
💡 What:
Implemented a static array cache for `content_url()` calls in the CSS and JS minifier classes (`includes/minify/class-css.php` and `includes/minify/class-js.php`), keyed by the current site's ID (`get_current_blog_id()`).

🎯 Why:
The minifier classes are instantiated multiple times (e.g., once for each inline script/style being extracted, or when replacing many image URLs in CSS via regex callbacks). `content_url()` is a WordPress core function that executes hooks and performs string manipulations. Calling it repeatedly inside these high-frequency paths creates unnecessary overhead.

📊 Impact:
Reduces redundant executions of the `plugins_url` and `content_url` WordPress hooks during asset minification, saving marginal processing time and memory allocations per asset replaced/minified on the page.

🔬 Measurement:
1. Enable CSS/JS minification in the plugin settings.
2. Load a page on the frontend with multiple CSS/JS files and inline styles.
3. Verify that styles and scripts are still minified correctly and URLs within CSS files (e.g., `background-image: url(...)`) resolve accurately without any performance degradation.

---
*PR created automatically by Jules for task [15855560451794430877](https://jules.google.com/task/15855560451794430877) started by @nilesh32236*

## Summary by CodeRabbit

* **Performance**
  * Improved content URL handling with multisite-aware caching.
  * Preserved compatibility by bypassing caching when URL filters are active.
  * Applied optimized URL resolution to generated cache and CSS image URLs.

* **Tests**
  * Added coverage for repeated paths, distinct paths, filtered URLs, and separate site caches.

* **Chores**
  * Updated internal caching guidance and repository ignore rules.

PR #541, opened 2026-08-06
⚡ Bolt: [Performance improvement] Centralize content_url caching into Util::cached_content_url

nilesh32236:
💡 What:
Replaces duplicated static array caching of `content_url` across `class-main.php`, `class-img-converter.php`, and `class-used-css.php` with a centralized `Util::cached_content_url()` utility method.

🎯 Why:
The blog-ID-keyed static cache arrays were previously duplicated in multiple minification and URL processing loops. Moving this logic to a single helper method removes duplicate code and ensures consistency with the existing base-URL caching pattern used in `class-main.php`, slightly improving memory usage and reducing redundancy when these operations occur.

📊 Impact:
Slight improvement in code maintainability and memory footprint by replacing inline static cache arrays initialized across multiple loops and callbacks with a central cached utility.

🔬 Measurement:
Run unit tests and `php -l` on the files changed to verify no regressions. Observe that caching features reliant on these URLs (minification, background image processing) remain functional when `wppo_settings` is enabled.

---
*PR created automatically by Jules for task [1076249866565627080](https://jules.google.com/task/1076249866565627080) started by @nilesh32236*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of cached image, CSS, and JavaScript asset URLs.
  * Enhanced compatibility for cached assets across multisite setups.
  * Preserved existing file path conversion and safety checks for image handling.

### DIFF
--- a/includes/class-img-converter.php
+++ b/includes/class-img-converter.php
@@ -904,20 +904,17 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Img_Converter' ) ) {
 				if ( empty( $local_path ) ) {
 					// If Util::get_local_path failed, manually resolve if it's a URL.
 					static $home_url    = array();
-					static $content_url = array();
+					$content_url_base   = untrailingslashit( Util::cached_content_url( '' ) );
 					$blog_id            = get_current_blog_id();
 
 					if ( ! isset( $home_url[ $blog_id ] ) ) {
 						$home_url[ $blog_id ] = untrailingslashit( home_url() );
 					}
-					if ( ! isset( $content_url[ $blog_id ] ) ) {
-						$content_url[ $blog_id ] = untrailingslashit( content_url() );
-					}
 
 					$local_base = wp_normalize_path( ABSPATH );
 
-					if ( strpos( $source_image, $content_url[ $blog_id ] ) === 0 ) {
-						$relative_path = substr( $source_image, strlen( $content_url[ $blog_id ] ) );
+					if ( strpos( $source_image, $content_url_base ) === 0 ) {
+						$relative_path = substr( $source_image, strlen( $content_url_base ) );
 						$local_base    = wp_normalize_path( WP_CONTENT_DIR );
 					} elseif ( strpos( $source_image, $home_url[ $blog_id ] ) === 0 ) {
 						$relative_path = substr( $source_image, strlen( $home_url[ $blog_id ] ) );
--- a/includes/class-img-converter.php
+++ b/includes/class-img-converter.php
@@ -903,9 +903,9 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Img_Converter' ) ) {
 
 				if ( empty( $local_path ) ) {
 					// If Util::get_local_path failed, manually resolve if it's a URL.
-					static $home_url    = array();
-					$content_url_base   = untrailingslashit( Util::cached_content_url( '' ) );
-					$blog_id            = get_current_blog_id();
+					static $home_url  = array();
+					$content_url_base = untrailingslashit( Util::cached_content_url( '' ) );
+					$blog_id          = get_current_blog_id();
 
 					if ( ! isset( $home_url[ $blog_id ] ) ) {
 						$home_url[ $blog_id ] = untrailingslashit( home_url() );
--- a/includes/class-main.php
+++ b/includes/class-main.php
@@ -2151,17 +2151,7 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Main' ) ) {
 
 			if ( $cached_file ) {
 				$basename = basename( $cached_file );
-				if ( false !== has_filter( 'content_url' ) ) {
-					$content_url = content_url( 'cache/wppo/min/css/' . $basename );
-				} else {
-					static $content_url_base = array();
-					$blog_id                 = get_current_blog_id();
-
-					if ( ! isset( $content_url_base[ $blog_id ] ) ) {
-						$content_url_base[ $blog_id ] = content_url( 'cache/wppo/min/css/' );
-					}
-					$content_url = $content_url_base[ $blog_id ] . $basename;
-				}
+				$content_url = Util::cached_content_url( 'cache/wppo/min/css/' . $basename );
 
 				$file_version = filemtime( Util::get_local_path( $cached_file ) );
 				$new_href     = $content_url . '?ver=' . $file_version;
@@ -2212,17 +2202,7 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Main' ) ) {
 
 			if ( $cached_file ) {
 				$basename = basename( $cached_file );
-				if ( false !== has_filter( 'content_url' ) ) {
-					$content_url = content_url( 'cache/wppo/min/js/' . $basename );
-				} else {
-					static $content_url_base = array();
-					$blog_id                 = get_current_blog_id();
-
-					if ( ! isset( $content_url_base[ $blog_id ] ) ) {
-						$content_url_base[ $blog_id ] = content_url( 'cache/wppo/min/js/' );
-					}
-					$content_url = $content_url_base[ $blog_id ] . $basename;
-				}
+				$content_url = Util::cached_content_url( 'cache/wppo/min/js/' . $basename );
 
 				$file_version = filemtime( Util::get_local_path( $cached_file ) );
 
--- a/includes/class-main.php
+++ b/includes/class-main.php
@@ -2150,7 +2150,7 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Main' ) ) {
 			$cached_file  = $css_minifier->minify();
 
 			if ( $cached_file ) {
-				$basename = basename( $cached_file );
+				$basename    = basename( $cached_file );
 				$content_url = Util::cached_content_url( 'cache/wppo/min/css/' . $basename );
 
 				$file_version = filemtime( Util::get_local_path( $cached_file ) );
@@ -2201,7 +2201,7 @@ if ( ! class_exists( 'PerformanceOptimise\Inc\Main' ) ) {
 			$cached_file = $js_minifier->minify();
 
 			if ( $cached_file ) {
-				$basename = basename( $cached_file );
+				$basename    = basename( $cached_file );
 				$content_url = Util::cached_content_url( 'cache/wppo/min/js/' . $basename );
 
 				$file_version = filemtime( Util::get_local_path( $cached_file ) );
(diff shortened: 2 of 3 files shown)


nilesh32236 (commented):
## Executive Summary

**Purpose:** Replaces three duplicated blog-ID-keyed static-array caches of content_url in class-main.php, class-img-converter.php and class-used-css.php with the existing, already-tested Util::cached_content_url() helper.

**Risk:** 🟢 LOW — Pure refactor to a pre-existing, unit-tested utility method (already used by the minify classes). Each new call site is behaviorally equivalent to the removed inline code: the has_filter('content_url') bypass and per-blog caching semantics are preserved, and php -l passes on all four touched files.

---

## MR Review Summary

A clean, behavior-preserving deduplication: the three call sites now delegate to Util::cached_content_url(), which was already introduced and tested for the minify classes. I verified each replacement is equivalent — same has_filter bypass, same per-blog caching, same untrailingslashit handling in the image converter — and confirmed no duplicated content_url caching patterns remain anywhere in the codebase. No blocking issues found.

**Ready to merge?** true

**Reasoning:** All three replacements are behavior-preserving calls into an existing, unit-tested utility; there are no correctness, security, or regression risks and no remaining duplicated patterns. Only a negligible, low-confidence note about per-path vs per-blog cache granularity.

### Strengths

- **includes/class-main.php:2154** — The refactor correctly preserves the subtle has_filter('content_url') bypass semantics that were in the inline code, routing to Util::cached_content_url() which mirrors the exact same check, so minified CSS/JS URLs behave identically whether or not a content_url filter is registered.
- **includes/class-img-converter.php:907** — Switching to Util::cached_content_url('') correctly preserves the previous untrailingslashit(content_url()) base with per-blog caching, while gaining consistency with the rest of the codebase. Bonus: when a content_url filter is present, the fallback now gets fresh (non-stale) values instead of the previously cached-first-call value.
- **includes/class-used-css.php:1129** — This call site gains caching it didn't have before (content_url() was invoked per stylesheet in the removal-URL loop), so the refactor is a small performance win here, not just a deduplication.

### Issues

- 🔵 **MINOR:** `includes/class-util.php:318` — The PR description claims a 'slightly improving memory footprint', but in the minify_css/minify_js callers the new per-path cache key stores one entry per unique minified basename per blog for the request lifetime, whereas the removed inline code stored a single base URL per blog and concatenated the basename. Impact is negligible (per-request static, and WP's own content_url() already statically caches internally), so this is informational only. **[low confidence]**
  > 💡 **How to fix:** If the per-path cache growth is ever a concern, cache the base content_url per blog (as the old code did) and concatenate $path; otherwise keep as-is since the current implementation is correct and the growth is trivial.

includes/class-img-converter.php
```diff
@@ -903,21 +903,18 @@ public static function get_img_path( string $source_image, string $format = 'web
 
 				if ( empty( $local_path ) ) {
 					// If Util::get_local_path failed, manually resolve if it's a URL.
-					static $home_url    = array();
-					static $content_url = array();
-					$blog_id            = get_current_blog_id();
+					static $home_url  = array();
+					$content_url_base = untrailingslashit( Util::cached_content_url( '' ) );
```
nilesh32236, line 907: 🔵 **MINOR**: The equals signs of the newly added assignment block are not aligned: `$content_url_base =` places `=` one column earlier than `$blog_id          =` and `static $home_url  =`. This matches the unresolved CodeRabbit spacing thread (the thread's specific figures of 3/12 spaces do not match the current bytes, but the underlying misalignment is real). Note that `composer lint` (phpcs, WordPress standard) currently passes this file, so this is cosmetic only and not a CI blocker.

> 💡 **How to fix:** Add one space after `$content_url_base` so its `=` aligns with the surrounding assignments: `$content_base_url  = untrailingslashit( Util::cached_content_url( '' ) );`.

```suggestion
$content_url_base  = untrailingslashit( Util::cached_content_url( '' ) );
```

nilesh32236 (commented):
## Executive Summary

**Purpose:** Centralizes the duplicated blog-ID-keyed static array caching of content_url in class-main.php, class-img-converter.php, and class-used-css.php into a single Util::cached_content_url() helper.

**Risk:** 🟢 LOW — A mechanical refactor that preserves the prior content_url caching semantics (the has_filter('content_url') fast-path and per-blog-ID keying) at all three call sites. No functional or security change; the project lint gate (composer lint) passes and the PHPUnit failures present on the PR base are unchanged by this PR.

---

## MR Review Summary

A clean, well-scoped refactor. Util::cached_content_url() is documented, blog-ID-keyed, and preserves the old has_filter('content_url') fast-path so output is identical at all three call sites. I verified content_url('') equals the previous content_url() used in class-img-converter.php, that composer lint (phpcs / WordPress standard) passes on all changed files, and that the ImgConverterTest / InlineCssTest / UtilCachedContentUrlTest failures all pre-exist on the PR base (859a49f4). Remaining findings are cosmetic: an off-by-one equals-alignment newly introduced at class-img-converter.php:907, and a pre-existing order-dependent unit test that leaks static cache state across test methods.

**Ready to merge?** true

**Reasoning:** The core changes are minimal and behavior-preserving, and the only in-scope finding is trivial whitespace that the actual lint gate does not even flag. No regressions were introduced (verified identical pre-existing test failures on the base commit).

### Strengths

- **includes/class-util.php:307** — Util::cached_content_url() is well designed: it preserves the old has_filter('content_url') fast-path to avoid caching context-dependent filtered output, and keys results by blog ID and path, matching the existing base-URL caching convention. The docblock clearly documents the semantics.
- **includes/class-main.php:2154** — The minify_css and minify_js call sites now fetch the cached URL directly with distinct output: caching the full content_url(path+basename) per site is functionally identical to the old base-URL plus basename split, so produced asset URLs are unchanged.
- **tests/php/UtilCachedContentUrlTest.php:38** — Good test coverage for the helper: repeated-call caching, independent path caching, filter-disables-caching, and per-blog-ID keying under switch_to_blog() — all behaviors the minification call sites depend on.

### Issues

- 🔵 **MINOR:** `tests/php/UtilCachedContentUrlTest.php:118` — Pre-existing and not introduced by this PR: test_results_are_keyed_per_blog_id fails (asserting 1 === 2 at line 118) when the suite runs in order, because Util::cached_content_url's PHP static `$cache` is populated by earlier test methods and cannot be reset by Brain Monkey. The test passes in isolation, so it is order-dependent and can give false confidence in the per-blog-ID behavior. Reported since it exercises the exact helper this PR centralizes on.
  > 💡 **How to fix:** Give each test an empty cache, e.g. run this test in a separate process, or refactor the helper so its cache is a class property that a test can reset via reflection or a test-only reset method.

nilesh32236 (commented):
## Executive Summary

**Purpose:** Replaces three duplicated blog-ID-keyed static content_url() caching blocks in class-main.php, class-img-converter.php, and class-used-css.php with calls to the centralized Util::cached_content_url() helper, reducing code duplication in minification and URL-resolution loops.

**Risk:** 🟢 LOW — The refactor is behavior-preserving: has_filter gating, blog-ID keying, and basename handling all match the prior inline logic (verified line-by-line against class-util.php:307), and phpcs passes. Only minor maintainability/memory nuance differs from the PR's stated impact.

---

## MR Review Summary

A sound, largely mechanical DRY refactor that correctly reuses the existing Util::cached_content_url() (established in PR #537) at four call sites. The three diffs are behavior-equivalent for the normal (no content_url filter) and filtered paths, and PHPCS is clean. Only minor caveats remain around the PR's memory-usage claim and a small per-path caching nuance.

**Ready to merge?** true

**Reasoning:** All four call sites produce output identical to the prior inline caching when no content_url filter is present, and correctly bypass caching when one is registered. No functional or security regressions found; only minor/low-confidence polish items remain.

### Strengths

- **includes/class-main.php:2154** — Correctly preserves the has_filter('content_url') gating from the original code by routing through Util::cached_content_url(), which re-checks the filter on every call instead of caching unconditionally. This keeps context-dependent filter output fresh while still caching when safe.
- **includes/class-img-converter.php:907** — Dropping the now-redundant per-file static $content_url array and delegating to the blog-ID-keyed helper removes a duplicated cache without changing the semantics of untrailingslashit() applied at the call site. Output is byte-identical to the prior inline cache.

### Issues

- 🔵 **MINOR:** `includes/class-util.php:318` — The helper caches the full content_url per unique path string, whereas the code it replaces in class-main.php cached a single base URL per blog ID and concatenated the basename. For N distinct minified assets this now stores N full URL strings (and N concatenated key strings) in a request-lifetime static, so the PR's stated impact of 'slightly improving memory requirement' is inaccurate for class-main.php: it is at best equal and marginally larger with many assets. The remaining redundancy-reduction benefit is definitive and warranted, only the memory claim should be reworded/appended to the changelog description. **[medium confidence]**
  > 💡 **How to fix:** Adjust the PR description/changelog to claim 'reduces duplicate code' rather than 'improves memory,' and consider scoring the helper by blog ID + a base-path so minifier call sites can prepend basenames, matching the LOWER memory profile of the original base-URL cache (e.g. cached_content_url_base() returning the base, then the caller appends basename).
- 🔵 **MINOR:** `includes/class-img-converter.php:907` — Minor behavior change when a content_url filter is registered: the prior inline cache stored untrailingslashit(content_url()) once per blog ID for the whole request, while the new path routes through cached_content_url(''), whose has_filter('content_url') gate disables caching so content_url('') is re-invoked per image processed. For a deterministic filter (typical CDN rewrite) this is a no-op and actually more correct; only a request-context-dependent filter would now produce a fresh base on every image, which is an edge case. **[low confidence]**
  > 💡 **How to fix:** No code change strictly required. If identical-per-request behavior is desired under a content_url filter, fall back to an unconditionally cached static base in the caller as before; otherwise document that per-image resolution under a registered filter is intentional.
- 🔵 **MINOR:** `tests/php/UtilCachedContentUrlTest.php:38` — Test gap: this PR adds no tests for the changed call sites; only the helper Util::cached_content_url() (already covered here) has coverage. The four updated call sites in get_img_path(), minify_css(), minify_js(), and inject_used_css() exercise the has_filter-present and blog-ID keying paths that would catch a regression in how the helper is invoked, but are not directly asserted. **[low confidence]**
  > 💡 **How to fix:** Add thin unit tests (or annotate existing behavior) asserting that minify_css()/minify_js() and inject_used_css() build URLs via cached_content_url() and that get_img_path() resolves URLs correctly when content_url() returns a base with and without a content_url filter.

includes/class-img-converter.php
nilesh-32236: Verified: the current PR head (`d5a6088`) already has all three `=` signs aligned at the same column — `static $home_url` (16 chars + 2 spaces), `$content_url_base` (17 chars + 1 space), `$blog_id` (8 chars + 10 spaces) all land on col 19. `composer lint` (WPCS, WordPress standard) passes on the file. The suggested `$content_url_base  =` (two spaces) would actually push its `=` to col 20 and misalign it. No code change needed.

PR #585, opened 2026-08-07
fix: guard zero-quality fallback and make PHPUnit suite green on PHP 8.5

nilesh-32236:
## Summary

Two changes bundled into one PR:

1. **Production fix**: `Img_Converter::resolve_encode_quality()` now rejects a `0` quality returned by `wp_image_quality()` (previously only `null` was guarded). A corrupted/missing core value now falls through to the plugin's fallback instead of encoding at quality `0`.

2. **Test infrastructure**: The PHPUnit suite was fatally broken on PHP 8.5 (Brain Monkey 2.x + Patchwork). This PR fixes the root causes and makes `composer test` pass (156 tests, exit 0).

## Test infra root-cause fixes

- **`patchwork.json`**: trimmed `redefinable-internals` to what Patchwork actually supports (`function_exists`, `getimagesize`). The removed `file_exists` / `is_readable` / `is_file` / `idn_to_ascii` entries caused fatal `Call to undefined function Patchwork\Redefinitions\*()` crashes.
- **`bootstrap.php`**: loads `templates/object-cache.php` before any test can eval-declare `wp_cache_*` stubs, and provides a default `$GLOBALS['wp_object_cache']` miss-on-read store. This fixes `Cannot redeclare function wp_cache_set()` and the `Call to a member function get() on null` cascade.
- **Shared test trait**: pre-registers common WP function stubs so per-test `when()` calls reconfigure the existing stub instead of re-declaring it across test classes sharing one process.
- **`InlineCssTest`**: uses the real object cache instead of redundant `wp_cache_*` stubs (which caused Patchwork `DefinedTooEarly`), fixes `function_exists` expectation ordering, stubs `wp_maybe_inline_styles` directly, creates missing fixture directories, and fixes the `content_url` stub to append the min-cache path.
- **`ImgConverterTest`**: invokes the non-static `resolve_encode_quality()` with a converter instance (was `ReflectionMethod::invoke(null, …)`), and aligns expectations with the current `(mime, fallback, size)` signature.
- **`ImageOptimisationTest`**: builds real LRU fixture files under ABSPATH, since `file_exists` / `is_file` are no longer Patchwork-redefinable.
- **`CacheTest` / `MainBlockAssetsTest`**: use trait stubs for `wp_maybe_inline_styles` / `get_bloginfo`; removed the now-unneeded `tests/php/stubs/minify-css-functions.php`.

## Verification

- `composer lint` — clean (exit 0)
- `composer test` — 156 tests, exit 0
- `npm test` — 255 passed
- `npm run lint:js` — 0 errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image-conversion quality handling by rejecting invalid or zero quality values and using the configured fallback instead.
  * Preserved consistent image optimization behavior across supported WordPress quality APIs.

* **Tests**
  * Strengthened image, cache, and inline CSS test coverage.
  * Improved test reliability through temporary file cleanup and more consistent WordPress environment setup.

* **Documentation**
  * Corrected PHPDoc alignment and formatting for generated configuration comments.


nilesh32236 (commented):
## Executive Summary

**Purpose:** Guards Img_Converter::resolve_encode_quality() against a zero quality value returned by wp_image_quality()/wp_get_image_encode_quality() (previously only null was rejected) so image conversion falls back to the plugin default instead of encoding at quality 0; and reworks the PHPUnit bootstrap/shared stubs (loading the object-cache drop-in early, trimming patchwork.json, pre-registering common WP stubs in a shared trait) so the suite passes on PHP 8.5 with Brain Monkey 2.x/Patchwork.

**Risk:** 🟢 LOW — The production change is a small, correct guard that only widens the fallback path and does not loosen security, auth, or data-loss protections; the bulk of the diff is isolated to test infrastructure. Residual impact is limited to an undocumented fall-through behavior change in the wp_get_image_encode_quality() branch and minor test-fixture hygiene gaps, plus a missing regression test for the headline zero-guard.

**Breaking Changes:**
- ⚠️ resolve_encode_quality() may now fall through from wp_get_image_encode_quality() to wp_image_quality() instead of returning the configured plugin fallback immediately when the size-aware helper returns an invalid (0/null) value

---

## MR Review Summary

A solid, well-scoped PR. The production fix is minimal and correct: a 0/null from both core quality helpers now consistently falls back to the plugin default instead of encoding at quality 0, and the early-return refactor lets a valid flat wp_image_quality() value satisfy the size-aware branch. The test-infrastructure overhaul is carefully reasoned and root-caused correctly for the Brain Monkey 2.x + Patchwork 'Cannot redeclare' and null-object-cache failures on PHP 8.5. The one blocking gap is that the headline zero-guard has no dedicated regression test; the remaining items are minor behavior/documentation and fixture-hygiene nits.

**Ready to merge?** false

**Reasoning:** The production guard is correct and the PHP 8.5 test-infra fixes are sound, but the exact behavior the PR is titled for (a core quality API returning 0 instead of null) has no regression test — existing resolve_encode_quality tests cover only null returns. Add a zero-guard test before merging; the remaining minor hygiene items are non-blocking.

### Strengths

- **includes/class-img-converter.php:194** — Correct and minimal guard: adding $quality > 0 to the existing null check in the wp_image_quality() branch prevents (int) 0 from being encoded at quality 0, and refactoring the size-aware branch into an equivalent early if/return lets a null/0 result also be satisfied by the flat wp_image_quality() API before reaching the fallback — an improvement over the old short-circuit-to-fallback path.
- **tests/php/bootstrap.php:31** — Loading the real templates/object-cache.php drop-in at bootstrap before any test can eval-declare wp_cache_* stubs, then providing a default miss-on-read $GLOBALS['wp_object_cache'] store and pre-registering common WP function stubs via the shared trait, elegantly sidesteps the Brain Monkey eval-declare 'Cannot redeclare' and Patchwork DefinedTooEarly traps, with clear explanatory comments.
- **tests/php/ImageOptimisationTest.php:573** — LRU fixture files are created, used, unlinked, and the empty parent dir removed so the test does not leak artifacts into the shared ABSPATH — the pattern the other newly-added file fixtures in this PR should follow.

### Issues

- 🔵 **MINOR:** `tests/php/ImgConverterTest.php:497` — No resolve_encode_quality() test exercises the behavior this PR adds: a zero quality from wp_image_quality() (or wp_get_image_encode_quality()). Existing tests cover only null (guards_null_size_aware_result at 637; falls_back_to_default at 574). The $quality > 0 guard at includes/class-img-converter.php:194 is the headline fix and is currently unprotected, so a future refactor could silently reintroduce the quality-0 encoding bug.
  > 💡 **How to fix:** Add a regression test asserting that when wp_get_image_encode_quality() returns null and wp_image_quality() returns 0, resolve_encode_quality() returns the supplied fallback rather than 0.

```suggestion
public function test_resolve_encode_quality_guards_zero_flat_quality(): void {
		Functions\when( 'wp_get_image_encode_quality' )->justReturn( null );
		Functions\when( 'wp_image_quality' )->justReturn( 0 );

		$converter  = $this->make_converter();
		$reflection = new ReflectionMethod( Img_Converter::class, 'resolve_encode_quality' );
		$reflection->setAccessible( true );

		$this->assertSame( 82, $reflection->invoke( $converter, 'image/avif', 82 ) );
	}
```
</details>
- 🔵 **MINOR:** `includes/class-img-converter.php:187` — Behavior change not described in the PR (which only mentions guarding wp_image_quality()'s zero): previously an invalid (0/null) result from wp_get_image_encode_quality() returned the plugin's configured $fallback immediately; now the code falls through and may return a value from wp_image_quality() instead, which can differ from the configured fallback on WP 7.1+ when the per-size helper returns invalid but the plain helper returns a valid core value. In practice both derive from the same core quality options so this is usually a no-op or an improvement, but it silently changes which value wins. **[low confidence]**
  > 💡 **How to fix:** Document this intentional fall-through in the method comment and PR description, or preserve the original strict-fallback semantics if the goal was only to reject zero, not to change resolution order.
- 🔵 **MINOR:** `tests/php/InlineCssTest.php:45` — InlineCssTest uses the WPPO_Test_Bootstrap trait, whose setUp() (bootstrap.php) already calls \Brain\Monkey\setUp() and pre-registers the common WP stubs. Calling \Brain\Monkey\setUp() a second time and re-registering the identical stub block duplicates the trait, resets the trait-registered stubs, and needlessly runs the setUp/teardown bridge twice (doubling Patchwork/error-handler setup). Harmless today only because the class re-registers what it needs, but fragile if the trait's stub list evolves. **[medium confidence]**
  > 💡 **How to fix:** Remove the second \Brain\Monkey\setUp() call and the stub registrations already provided by the trait, keeping only the overrides that differ (is_user_logged_in, apply_filters); if the extra call is genuinely needed, document why.
- 🔵 **MINOR:** `tests/php/InlineCssTest.php:283` — The theme fixture directory /tmp/wordpress/wp-content/themes/t and its style.css, plus the min-cache css files under /tmp/wordpress/wp-content/cache/..., are created but never removed. tearDown() only unlinks small_css_file and large_css_file, and adjacent suites depend on @rmdir() silently failing. Leftover files in a shared ABSPATH can leak across CI runs and make file_exists()/is_file() assertions order-dependent, unlike ImageOptimisationTest which cleans its fixtures.
  > 💡 **How to fix:** Track the created source/cache files and remove them plus the created directories in tearDown(), mirroring ImageOptimisationTest.php:573-579.
- 🔵 **MINOR:** `tests/php/bootstrap.php:179` — The shared content_url() stub uses returnArg(), which returns null when core code calls content_url() with no argument (the most common call). A lossy default can silently collapse content-path/URL building in tests that do not override it; only InlineCssTest supplies a correct alias that appends the base and path. **[low confidence]**
  > 💡 **How to fix:** Stub content_url() to append the argument to a base URL (e.g. return 'http://example.com/wp-content' . (string) $path), consistent with the override InlineCssTest already applies.
- 🔵 **MINOR:** `tests/php/ImageOptimisationTest.php:570` — The temp fixture cleanup (unlink loop + @rmdir) only runs after the assertions pass. If an assertion throws mid-test, the 101 lru-*.jpg files and the uploads directory leak under /tmp/wordpress/wp-content/uploads, and leftovers could affect subsequent runs in the same sandbox. **[medium confidence]**
  > 💡 **How to fix:** Move the cleanup to a finally block or to tearDown() so files are removed even on assertion failure.
- 🔵 **MINOR:** `tests/php/bootstrap.php:49` — The default anonymous $GLOBALS['wp_object_cache'] store only implements get() and set(). The real drop-in helpers wp_cache_add(), wp_cache_delete(), wp_cache_flush() and wp_cache_add_salt() call $wp_object_cache->method() directly, so any regression-added plugin path calling those through the real cache helpers would fatal with 'Call to undefined method' rather than fail cleanly. Acceptable today because ObjectCacheTest overrides it, but brittle. **[low confidence]**
  > 💡 **How to fix:** Implement the full WP_Object_Cache surface used by the drop-in (add/delete/flush/add_salt) as no-ops/false-returning on the anonymous class so future paths fail gracefully instead of fataling.
- 🔵 **MINOR:** `tests/php/bootstrap.php:84` — The warning suppressor no longer checks the error type (E_WARNING / E_USER_WARNING), so any error message merely containing 'Patchwork Redefinitions' is swallowed. Now that patchwork.json has been trimmed, those Patchwork Redefinitions calls arguably should not occur at all — this handler may be masking the real symptom. **[low confidence]**
  > 💡 **How to fix:** After trimming patchwork.json, confirm no Patchwork Redefinitions warnings remain on PHP 8.5 and either delete the handler or keep an errno-guard; if none remain the blanket suppression is no longer needed.

nilesh-32236:
Addressed the review feedback in `6c609eff`:

**Blocking**
- ✅ **Missing zero-guard regression test** — added `test_resolve_encode_quality_guards_zero_flat_quality()` asserting that when `wp_get_image_encode_quality()` returns null and `wp_image_quality()` returns `0`, `resolve_encode_quality()` returns the supplied fallback instead of `0`.

**Minor**
- ✅ **Fall-through behavior undocumented** — documented the size-aware → flat-helper fall-through in the `resolve_encode_quality()` docblock (and fixed a stray duplicate `/**` opener).
- ✅ **`content_url()` returnArg() lossy default** — the shared trait stub now appends the path to the base URL, matching the InlineCssTest override.
- ✅ **InlineCssTest fixture hygiene** — added a `tearDown()` safety net that removes any left-over theme/cache fixture files and dirs, and made the minify tests create their cache directory so they're self-contained.
- ✅ **ImageOptimisationTest cleanup on failure** — LRU fixture cleanup moved into a `finally` block so files are removed even if an assertion throws.
- ✅ **`$wp_object_cache` full surface** — added `add`/`delete`/`flush`/`add_salt` no-ops so future cache-backed paths fail gracefully instead of fataling.
- ⏭️ **InlineCssTest setUp duplication** — not changed: InlineCssTest defines its own `setUp()`, which *overrides* the trait's (PHP trait resolution), so the trait's stub registrations are not applied to this class; the re-registration is required, not duplicated.
- ⏭️ **Error-handler errno guard** — reverting to the guarded version broke a test on PHP 8.5 (Patchwork `Redefinitions` notices surface as `E_USER_ERROR` during shutdown), so the unguarded suppression is intentionally retained.

Verification: `composer test` (157 tests, exit 0), `composer lint` (exit 0).

tests/php/bootstrap.php
```diff
@@ -30,11 +30,101 @@
...
+if ( ! function_exists( 'wp_cache_set' ) ) {
+	require_once __DIR__ . '/../../templates/object-cache.php';
+}
+
+// Provide a default in-memory object cache so the real salted cache helpers
+// (which read $GLOBALS['wp_object_cache']) do not fatal when tests touch the
+// plugin's cache-backed code paths. ObjectCacheTest replaces this with its own
+// richer stub in setUp(); every other test gets a miss-on-read store.
+// phpcs:disable WordPress.WP.GlobalVariablesOverride.Prohibited
+if ( ! isset( $GLOBALS['wp_object_cache'] ) ) {
```
nilesh32236, line 48: 🔵 **MINOR**: The new default $GLOBALS['wp_object_cache'] anonymous-class store only implements get/set/add/delete/flush/add_salt. The just-loaded templates/object-cache.php drop-in delegates additional methods to the store (flush_group(), replace(), get_multiple()/set_multiple()/delete_multiple(), incr()/decr()), so any test touching those paths before ObjectCacheTest replaces the global will fatal with 'Call to undefined method'. The suite is green only because current tests never hit them — order- and coverage-dependent fragility that will bite future tests. **[medium confidence]**

> 💡 **How to fix:** Complete the fallback store so the drop-in's full surface is always present: add the missing methods (flush_group, replace, get_multiple, set_multiple, delete_multiple, incr, decr) looping over the existing get/set/delete primitives, or build a miss-on-read store from the real WP_Object_Cache.

```suggestion
public function flush_group( $group ) { unset( $this->cache[ $group ] ); return true; }
public function replace( $key, $value, $group = '' ) { if ( ! isset( $this->cache[ $group ][ $key ] ) ) { return false; } $this->cache[ $group ][ $key ] = $value; return true; }
public function get_multiple( $keys, $group = '' ) { $out = array(); foreach ( (array) $keys as $k ) { $out[ $k ] = $this->get( $k, $group ); } return $out; }
public function set_multiple( $items, $group = '' ) { foreach ( (array) $items as $k => $v ) { $this->set( $k, $v, $group ); } return true; }
public function delete_multiple( $keys, $group = '' ) { foreach ( (array) $keys as $k ) { $this->delete( $k, $group ); } return true; }
public function incr( $key, $offset = 1, $group = '' ) { $v = ( isset( $this->cache[ $group ][ $key ] ) ? (int) $this->cache[ $group ][ $key ] : 0 ) + (int) $offset; $this->cache[ $group ][ $key ] = $v; return $v; }
public function decr( $key, $offset = 1, $group = '' ) { return $this->incr( $key, - (int) $offset, $group ); }
```

tests/php/bootstrap.php
```diff
@@ -30,11 +30,101 @@
...
+	};
+}
+// phpcs:enable WordPress.WP.GlobalVariablesOverride.Prohibited
+
 // Suppress Patchwork internal redefinition warnings on PHP 8.5 during shutdown.
 // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_set_error_handler -- Suppress Patchwork warnings on PHP 8.5
 set_error_handler(
 	static function ( $errno, $errstr ) {
-		if ( ( E_USER_WARNING === $errno || E_WARNING === $errno ) && str_contains( $errstr, 'Patchwork\Redefinitions' ) ) {
+		if ( str_contains( $errstr, 'Patchwork\Redefinitions' ) ) {
```
nilesh32236, line 127: 🔵 **MINOR**: The Patchwork warning-suppression handler was broadened from E_USER_WARNING/E_WARNING only to every errno whose message contains 'Patchwork\Redefinitions', installed once at bootstrap and never restored. Because it swallows any such message, unrelated real warnings/notices mentioning that string during the whole suite could be silently masked, making genuine failures harder to diagnose. **[medium confidence]**

> 💡 **How to fix:** Keep the strict errno filter (E_USER_WARNING || E_WARNING) in addition to the message check, since the pre-existing crash it guards against was warning-level.

tests/php/bootstrap.php
```diff
@@ -102,6 +192,41 @@ trait WPPO_Test_Bootstrap {
 	protected function setUp(): void { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
 		parent::setUp();
 		\Brain\Monkey\setUp();
+
+		// Pre-register frequently used WP functions to avoid "Cannot redeclare"
```
nilesh32236, line 196: 🔵 **MINOR**: The shared-trait pre-registration of common stubs only runs for the ~7 test classes that do NOT define their own setUp(). Every class with a custom setUp — including exactly the three files changed in this PR (InlineCssTest.php:43, ImgConverterTest.php:37, ImageOptimisationTest.php:29) — overrides the trait method (class methods win over trait methods, and their parent::setUp() is TestCase's, not the trait's). The comment claims this guards all classes against 'Cannot redeclare', but those classes re-declare the same stubs inline, creating two sources of truth that will drift.

> 💡 **How to fix:** Extract the pre-registration into a shared helper (e.g. a trait registerCommonStubs()) and call it from both the trait's setUp and each custom setUp, then correct the comment to state which classes it actually covers.

tests/php/ImageOptimisationTest.php
```diff
@@ -527,9 +527,23 @@ public function test_generate_svg_base64_preserves_normal_dimensions(): void {
...
 		$this->stub_local_path_resolution();
-		Functions\when( 'file_exists' )->justReturn( true );
-		Functions\when( 'is_readable' )->justReturn( true );
-		Functions\when( 'is_file' )->justReturn( true );
+
+		// post_process_img_dimensions() guards on file_exists()/is_file() with
+		// the real filesystem (those functions are no longer Patchwork
+		// redefinable), so create the resolved files under ABSPATH that
+		// Util::get_local_path() maps each lru-N.jpg URL to.
+		$upload_dir = '/tmp/wordpress/wp-content/uploads';
```
nilesh32236, line 535: 🔵 **MINOR**: post_process_img_dimensions() keeps its LRU in a static local ($img_size_cache, class-image-optimisation.php:297) that survives across test runs in the same process. The fixture files use fixed paths (lru-0..100 under /tmp/wordpress/wp-content/uploads) and are deleted after the test, so any re-run of this test (e.g. --repeat) or a future test using the same URLs would hit stale cache entries and make the assertSame(101, $getimagesize_calls) count non-deterministic. **[low confidence]**

> 💡 **How to fix:** Use unique per-run fixture paths (e.g. a tempnam-based prefix) and, if the static cache ever becomes an issue, reset it via reflection, or add a comment documenting the in-process static-state assumption.

tests/php/bootstrap.php
```diff
@@ -102,6 +192,41 @@ trait WPPO_Test_Bootstrap {
...
+				}
+				return $path;
+			}
+		);
+		\Brain\Monkey\Functions\when( 'home_url' )->justReturn( 'http://example.com' );
+		\Brain\Monkey\Functions\when( 'get_current_blog_id' )->justReturn( 1 );
+		\Brain\Monkey\Functions\when( 'WP_Filesystem' )->justReturn( false );
+		\Brain\Monkey\Functions\when( 'sanitize_text_field' )->returnArg();
+		\Brain\Monkey\Functions\when( 'wp_unslash' )->returnArg();
+		\Brain\Monkey\Functions\when( 'content_url' )->alias(
```
nilesh32236, line 222: 🔵 **MINOR**: The content_url() stubs diverge from core in two places: InlineCssTest's Functions\when('content_url')->returnArg() returns an empty string for a no-arg call (InlineCssTest.php:75-79), and the trait's alias returns 'http://example.com/wp-content' . $path without inserting '/' for non-slash-prefixed paths (core adds it). The result can silently produce malformed URLs (e.g. content_url('style.css') -> 'http://example.com/wp-contentstyle.css') in future tests that forget to override the stub. **[low confidence]**

> 💡 **How to fix:** Unify on one core-mirroring alias used by both the trait and InlineCssTest: return the base URL and prepend '/' when the path is non-empty and does not already start with '/'.

```suggestion
static function ( $path = '' ) {
	$url = 'http://example.com/wp-content';
	if ( '' !== (string) $path && '/' !== (string) $path[0] ) {
		$url .= '/';
	}
	return $url . (string) $path;
}
```

tests/php/InlineCssTest.php
```diff
@@ -42,6 +42,7 @@ class InlineCssTest extends \PHPUnit\Framework\TestCase {
 	 */
 	protected function setUp(): void {
 		parent::setUp();
+		\Brain\Monkey\setUp();
```
nilesh32236, line 45: 🔵 **MINOR**: The PR added an explicit \Brain\Monkey\setUp() (and tearDown() at line 116) even though parent::setUp() already runs WPPO_Test_Bootstrap::setUp(), which calls \Brain\Monkey\setUp() and registers the shared stubs. The explicit calls are redundant (Brain Monkey setUp is idempotent), and the class's own stub block (lines 58-79) duplicates ~80% of the trait's list (bootstrap.php:196-229), so the two lists will drift as the suite evolves. **[medium confidence]**

> 💡 **How to fix:** Drop the explicit \Brain\Monkey\setUp()/tearDown() calls in InlineCssTest (the trait covers them) and keep only the overrides that genuinely differ, e.g. content_url() -> returnArg().

nilesh32236 (commented):
## Executive Summary

**Purpose:** Fixes a production bug in Img_Converter::resolve_encode_quality() where a zero/null quality returned by core image-quality helpers would encode at quality 0, and reworks the PHPUnit bootstrap/test infrastructure (Brain Monkey 2.x + Patchwork on PHP 8.5) so the suite runs green (157 tests).

**Risk:** 🟢 LOW — The production change is small, correctly guarded, covered by a dedicated regression test, and has no API/behavioral break for healthy core installs; the rest is test infrastructure, verified green on PHP 8.2 with clean phpcs. The single important finding (incomplete default object-cache store) is latent and only affects future or order-dependent test paths, not current behavior.

**Breaking Changes:**
- ⚠️ resolve_encode_quality() now falls through to wp_image_quality() when the WP 7.1+ size-aware helper returns null/0 instead of immediately returning the plugin fallback; on a corrupted size-aware helper a valid flat core value is now used in preference to 82.

---

## MR Review Summary

The zero-quality production fix in Img_Converter::resolve_encode_quality() is correct, well-documented, and pinned by a focused regression test, and the suite genuinely passes (157 tests, phpcs clean) in default order while being far more order-robust than before. The test-infrastructure overhaul works, but the new bootstrap default object-cache store is an incomplete stand-in and the shared-trait stub pre-registration is silently skipped by the very classes this PR modifies — fragility/maintainability concerns rather than current breakage.

**Ready to merge?** true

**Reasoning:** No critical or blocking-important defects: the production guard is sound and the full suite plus lint pass. The single important issue (incomplete default object-cache store) is latent and only affects future or order-dependent test paths, so it does not block merge; address it as a follow-up for robustness.

### Strengths

- **includes/class-img-converter.php:200** — resolve_encode_quality() correctly rejects null, 0, false, negative and non-numeric values from both the size-aware and flat core-helper branches, so a corrupted/missing core quality can no longer degrade to an encode at quality 0, and the fall-through behavior is explicitly documented in the PHPDoc.
- **tests/php/ImgConverterTest.php:656** — test_resolve_encode_quality_guards_zero_flat_quality() is a focused, isolated regression test that pins the exact headline fix (flat helper returning 0 falls back to the supplied default) with both avif and jpeg fallbacks.
- **tests/php/ImageOptimisationTest.php:579** — Using try/finally for fixture cleanup guarantees the 101 temp JPEGs and upload dir are removed even when an assertion fails — a good pattern the other rewritten tests would benefit from.
- **tests/php/bootstrap.php:39** — Loading the object-cache drop-in early, guarded by function_exists('wp_cache_set'), is a sound root-cause fix for the 'Cannot redeclare function wp_cache_set()' fatal, and ObjectCacheTest's require_once becomes a harmless no-op.
- **patchwork.json:1** — Trimming redefinable-internals to what Patchwork actually supports is the correct fix, and a grep confirms no remaining test stubs the removed internals (file_exists/is_file/is_readable/idn_to_ascii), so there are no silently-ignored mocks.

### Issues

- 🔵 **MINOR:** `includes/class-img-converter.php:336` — Pre-existing dead code in the very method this PR reworks: the second `if ( -1 === $quality )` block (lines 334-342) is unreachable because $quality is always resolved or set to 82 at lines 263-275, yet it calls self::resolve_encode_quality('image/avif', $size), passing the $size array into the int-typed $fallback parameter — a TypeError (verified on PHP 8.2) if the block is ever reached. The ordering coincidence that keeps it dead is fragile.
  > 💡 **How to fix:** Delete the unreachable block, or if it is meant to handle the 'both' path, correct the calls to $this->resolve_encode_quality( $mime, 82, $size ) to match the (string, int, array) contract.

```suggestion
if ( 'both' === $format ) {
				$avif_quality = $this->resolve_encode_quality( 'image/avif', 82, $size );
				$webp_quality = $this->resolve_encode_quality( 'image/webp', 82, $size );
			} else {
				$mime    = in_array( $format, array( 'avif', 'both' ), true ) ? 'image/avif' : 'image/webp';
				$quality = $this->resolve_encode_quality( $mime, 82, $size );
			}
```
</details>
- 🔵 **MINOR:** `tests/php/InlineCssTest.php:332` — Pre-existing order-dependence that the trait does not fully eliminate: under --order-by=random, InlineCssTest::test_minify_css_minifies_third_party_path_data_handle fails with 'has_filter is not defined nor mocked' and ImageOptimisationTest::test_prioritize_lcp_in_buffer_graceful_without_html_api fails with 'is_singular is not defined nor mocked' (both also failed on the base branch; the PR reduces random-order failures from 20 to 2). Adding these to the trait's when() list does NOT fix it (verified empirically).
  > 💡 **How to fix:** Stub has_filter (-> justReturn(false)) in the minify test and is_singular (-> justReturn(false)) in test_prioritize_lcp_in_buffer_graceful_without_html_api so both are self-contained and pass regardless of test order.

